PR #3, opened 2023-04-05
Add "Users should never see an address document". This is not a stand…

rt121212121:
…ard.

### This pull request:

contains a document and was solicited for inclusion by Deggen

### DIFF
--- a/scripts/0049.md
+++ b/scripts/0049.md
@@ -1,0 +1,81 @@
+# BRC-49: Users should never see an address
+
+## Abstract
+
+Addresses should have been a temporary measure for Bitcoin, whether the original Bitcoin SV or the forked off Bitcoin Core. Something that bridges the gap between the initial release and a user friendly next step. Unfortunately, for lack of any better idea they were adopted as a central feature and canonised alongside a limited set of allowed of standard scripts. Worse still is they were intended to be phased out, but the replacement of Paymail is a flawed system that never met it's promise and is inherently insecure and broken.
+
+Let's work our way through from Bitcoin addresses, to standard scripts, to Paymail and finally to examining what we should be using instead.
+
+## Motivation
+
+This document is intended to look at addresses and identify why we use them and to suggest that they are unnecessary and should be phased out.
+
+## The Bitcoin address
+
+This address encodes two pieces of information, an indication of what type of script template should be used and the extracted part of the script template that differs. The supported script templates are either P2PKH, which is still the most commonly used template, or P2SH with which can no longer be used to store funds. And the differing part is a 160-bit hash of either the public key (P2PKH) or the real script that will be provided when any funds locked up in the script are spent (P2SH). The representation of the address is called base58, and was intended to represent the output script in a way where a user could transcribe it without confusing which characters they were transcribing.
+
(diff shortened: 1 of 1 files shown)


ty-everett:
While this is a well-written and thought-provoking piece, I am concerned about the scope and whether it belongs in this repository in its current form. My understanding is that this repository is for templates, standards and protocols. This document doesn't make any specific technical proposal, beyond that addresses should not be used. I think that this document, if distilled and refined into a summary, would make an excellent "Motivations" section for a technical standard that proposes a specific replacement for addresses, or defines a new and specific way of addressing the problems raised. Such a standard would make it clear what someone would need to do in order to be considered "in compliance" with the document, ideally with either links to reference implementations or test vectors.

The document raises some nebulous but very important and valid points about what the future _might_ be, but it doesn't go so far as to propose anything specific in a concrete way. What are the specific requirements or criteria that someone would need to follow in order to be compliant with BRC-49? If the stipulations and specifications are simply that "wallets who follow this specification will never display addresses or paymail handles", my suggestion would be to re-structure your document in the form of a standard for them to follow, for example:

**Title**: Removal of Addresses and Paymail Handles from Wallets
**Abstract**: This standard proposes that wallets remove the display of addresses in their user interfaces... <specific proposals for wallets>
**Motivations**: <your arguments for why wallets should do this>
**Specification**: <specific things that wallets must do in order to comply with your requirements>
**Implementation**: <links to examples of compliant wallets>
**References**: Links to your previous writings about this that support your arguments, as referenced throughout your document, such as your Paymail article on CoinGeek

If the purpose of this document was merely to start a discussion for future standards related to this issue (a discussion which I would support and want to be a part of), then I think it would be better to open an issue in the repository, providing this write-up as a starting post in the discussion so that people can talk it over and come up with specific solutions. It would most certainly be beneficial to then formalize and memorialize anything that come out of that discussion as one or multiple BRCs.

> Finally, a meta-note on process: I don't feel in this case that I can or should act unilaterally in making the decision on this document. I think anything should be agreed by everyone involved, as it will certainly set some precedent for the future.

rt121212121:
The purpose of this repository is so people can share protocols and concepts even if others haven't approved them. There is not as I understand it supposed to be gate-keeping. If in order to get a document included it needs to meet the approval of an unnamed panel who have to add their two cents, then suddenly it becomes TSC v2.0. Additionally, I would note your own protocols went straight in without the panels of approvers being formed to change them suitably, so you'd have to back those out.

sirdeggen:
After reading through, I don't see why not to include it. It's important we allow dissenting opinions and not filter so long as the BRC is sincere which this certainly is.

"beyond that addresses should not be used" <= that's the technical standard being proposed here I think. It should be put forward, and there ought to be somewhere we can continue the conversation, like an issue pointing to this BRC as reference.

ty-everett:
Yes, this makes sense and thank you for the clarification. I don't want to be an editor and people should merge things they feel are relevant. I will be less formal with this process going forward.

rt121212121:
I would add a final note that I am fine with this being placed into an opinion folder and separated from the protocols.

sirdeggen:
I think they're all opinions. Some opinions become popular as references we can point back at with mutual understanding.

PR #5, opened 2023-04-05
Update 0014.md

sirdeggen:
- Add Buffer example of binary
- Correct a missing hex digit
- Add SAMS format


ty-everett (approved):
Good catch on the missing hex character, and the new format makes sense.

PR #34, opened 2023-04-25
Brayden working

BraydenLangley:
### This pull request:

### Summary

In this standard, we define the data structures, key derivation protocols, signing, and verification procedures for identity certificates. We also discuss how responsible certificate issuers can handle key compromises.


peer-to-peer/0052.md
```diff
@@ -0,0 +1,268 @@
...
+### Secure Verifier Field Revelation Protocol
+
+In order to preserve  data integrity we need a way to keep the fields encrypted such that only particular fields can be revealed as needed.
+We define a standard why to do this using a keyring structure as defined above.
+
+When a prover reveals the fields to a verifier, they will follow these steps:
+
+1. Retrieve the relevant symmetric field encryption keys
+2. Encrypt the keys for the verifier following the [BRC-2](../wallet/0002.md) process for counterparty encryption.
+3. Construct the keyring with the field revelation keys and their associated field names according to the structure defined above.
```
ty-everett, line 244:
```suggestion
In order to preserve data integrity, we need a way to keep certificate fields encrypted such that only particular fields can be revealed as needed.
We stipulate the use of a keyring structure for selective field revelation to verifiers. As the keyring is not included in the certificate's signature (only the encrypted field values themselves), verifiers can still check the authenticity of the certificate without access to all keyring keys.

When a prover reveals the fields to a verifier, they will follow these steps:

1. Retrieve the relevant symmetric field encryption keys
2. Encrypt the keys for the verifier following the [BRC-2](../wallet/0002.md) process for counterparty encryption.
3. Construct the keyring with the field revelation keys and their associated field names according to the structure defined above.

For step 2, we stipulate the use of the following BRC-43 attributes when deriving the encryption keys:

BRC-43 Attribute | Value
-----------------|----------------
Security Level   | `2`
Protocol ID      | `authrite certificate field encryption`
Key ID           | `xxxxx yyyyy`
Counterparty     | `self`

**Where:**
- `xxxxx` is the certificate Serial Number
- `yyyyy` is the field name from the certificate
```

ty-everett (approved):

PR #43, opened 2023-05-12
typos etc.

raspi-user:
### This pull request:

### Summary

Typos etc.


ty-everett:
Thanks and good catches, Bob. As we have discussed internally I am fine with the changes to my BRCs but would like to get consent from @rt121212121 before making changes to his document BRC-51. Roger, are these small updates OK by you?

rt121212121:
Looks like good improvements, I am not American so I use s instead of z and will continue to do so and want those to remain.

PR #45, opened 2023-05-13
Update README.md to remove the word Standard and reorder the flow of the document a little

sirdeggen:


sirdeggen:
@ty-everett don't want to just blast in here but these are my initial thoughts.

ty-everett:
Makes sense, what is `SUMMARY.md` and is it complete / ready for inclusion yet?

ty-everett:
Ah, just noticing this is still a draft. I like the changes to the README in moving the contribution sections above the long list of standards.

sirdeggen:
Summary is to control the homepage of the Gitbooks so that it will be something we design rather than just the first document in the root directory alphabetically.

ty-everett:
That makes sense. Once this is moved out of Draft stage then it can be merged

PR #48, opened 2023-07-11
BEEF Transactions & Compound Merkle Paths

sirdeggen:
### This pull request:

Proposes two new interrelated binary data formats for SPV

1. Compound Merkle Paths
2. BEEF Transactions

### Summary

Inviting discussion around a way to send only the minimum required data for SPV based on incremental improvement to existing standards. Combination of a few good ideas from others into a unified solution which should be compatible with multiple implementations and importantly needed for use in Merkle Service Nodes which are just starting to become a necessary component of the network thanks to Teranode.

### DIFF
--- a/transactions/0061.md
+++ b/transactions/0061.md
@@ -27,39 +27,44 @@ Compound Path
 ### Example
 
 ```javascript
-nLeafs // repeat for each path
-  height offset leaf...  // repeat for each leaf
-  height offset leaf...  // repeat for each leaf
-  height offset leaf...  // repeat for each leaf
-  height offset leaf...  // repeat for each leaf
-  height offset leaf...  // repeat for each leaf
+height offset leaf...  // repeat for each leaf
+height offset leaf...  // repeat for each leaf
+height offset leaf...  // repeat for each leaf
+height offset leaf...  // repeat for each leaf
+height offset leaf...  // repeat for each leaf
 // ...etc.
 ```
 
 ## Hex Example
 
 ```javascript
-05 // nLeafs VarInt
-// ----------------------
 02 // height 1 byte
-01 // offset VarInt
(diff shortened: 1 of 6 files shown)


transactions/0061.md
```diff
@@ -0,0 +1,174 @@
+# BRC-61: Compound Merkle Path Format
```
icellan, line 1: Is this at all related to the bode that has already been implemented in `go-bc`? https://github.com/libsv/go-bc/blob/master/merkleproof.go

transactions/0062.md
```diff
@@ -0,0 +1,240 @@
...
+- Newest Tx
+
+As soon as the Merkle Paths are receieved we can calculate the roots and lookup their blockheaders. If they're not valid then validation can stop there - rejected. Then we look at the oldest ancestor - if it's valid then its children can be validated, and so on until we reach the most recent tx.
+
+## Specification
+
+BEEF combines thinking from several formats into one binary stream, prefixed with a header for disambiguation.
+
+- Raw Transaction Format: [BRC-12](./0012.md)
+- Extended Format (EF): [BRC-30](./0012.md)
```
icellan, line 51: @sirdeggen Link is incorrect. Linking to 0012.md

transactions/0061.md
sirdeggen: The existing implementation is for individual proofs for one txid. This allows the storage of multiple without repeating internal hashes.

That's one distinction. The other is that there is no extraneous info. This would be the minimum data required to be able to validate. The existing proof formats all have extra data which is nice to have but not necessary.

sirdeggen: For example we could have paths for 1,000 transactions all in one block. The idea being a merkle service would be asked to share the paths for all of them in one api request.

The compound proof would reduce the number of internal branch hashes depending on the proximity of the transactions in the block. Rather than having to repeat hashes, it would only ever share a subset of the hashes required to cover every path.

PR #49, opened 2023-08-23
WIP: SPV Generic Reference

sirdeggen:
### This pull request:

### Summary

A more generic definition of SPV which doesn't refer to any specific implementation - as a clear reference definition.

The current BRC-9 is too implementation specific, while this should cover all known implementations, and more succinctly defines what would constitute SPV from a higher level.


sirdeggen:
@ty-everett please consider renaming BRC-9 something more specific to Babbage stack, I've tried to outline a non specific version here for the sake of explaining to other developers who are not familiar with your stack or SPV generally.

Let's discuss here before merging anything.

sirdeggen:
Your argument is valid, agreed. Adding to the summary page and then merging as is.

sirdeggen:
Updating to 67

PR #54, opened 2023-09-02
txid index inclusion

sirdeggen:
### This pull request:

### Summary

we need to encode the index of txid within the block in order to map single merkle paths from the format

### DIFF
--- a/transactions/0061.md
+++ b/transactions/0061.md
@@ -38,6 +38,10 @@ For each level of the merkle path the opposite hash from the one which can be ca
 > 3. `nLeaves` is repeated for each height, followed by the corresponding `offset` and `leaf` for each.
 
 
+## Important Inclusion
+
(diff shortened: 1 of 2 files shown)


transactions/0061.md
```diff
@@ -61,14 +68,20 @@ c58e40f22b9e9fcd05a09689a9b19e6e62dbfd3335c5253d09a7a7cd755d9a3c // 32 byte hash
 // ----------------------
 // implied end of leaves at this height
 // height of next leaves is therefore 0
-02 // nLeafs at this height VarInt
+04 // nLeafs at this height VarInt
+// ----------------------
+00 // offset VarInt
+8c00bb9360e93fb822c84b2e579fa4ce75c8378ae87f67730a49552f73c56ee8 // 32 byte hash
```
dorzepowski, line 74: Maybe it'll be good to mention here in comment that for example purpose this is the id of on of  transaction for which CMP is build

PR #55, opened 2023-09-15
Improvement/beef

sirdeggen:
### This pull request:

### Summary

Since BEEF standard is as yet unimplemented, this substantial change is incorporated into the original proposal rather than made anew. The proposal is now that all transactions must be included in full, not with extended data only, since this prevents the receiver from determining whether the extended values are truly within the txid provided.

The intention is to allow use of this format without relying on a h(outpoint, script, sats) keyed lookup - which we expect to see in both overlay nodes and teranode, but not lite clients.

### DIFF
--- a/transactions/0062.md
+++ b/transactions/0062.md
@@ -1,9 +1,9 @@
 # BRC-62: Background Evaluation Extended Format (BEEF) Transactions
 
-Deggen (deggen@kschw.com)
-Damian Orzepowski (damian.orzepowski@4chain.studio)
-Wojciech Regulski (wojciech.regulski@4chain.studio)
-Arkadiusz Osowski (arkadiusz.osowski@4chain.studio)
+Deggen (deggen@kschw.com)  
+Damian Orzepowski (damian.orzepowski@4chain.studio)  
+Wojciech Regulski (wojciech.regulski@4chain.studio)  
+Arkadiusz Osowski (arkadiusz.osowski@4chain.studio)  
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


transactions/0062.md
```diff
@@ -179,62 +149,48 @@ console.log({ correctOrder: khanTopologicalSort(graph) })
...
+    a. RawTx bytes double sha256 to get the txid.
+    b. Store in hashmap from txid => parsedTx
+    c. If there is a Merkle path after the tx then we lookup the CMP using the CMP index number.
+        1. Lookup the txid within level 0 leaves of the CMP to get the index of the txid within a block.
+        2. Calculate the Merkle root with the index, txid, and CMP data.
+        3. Add the merkle root to an array which will be used in a request to our local header service once all transactions have been parsed.
+        4. Mark the tx as valid or not as soon as the header service responds.
+    d. Otherwise we run local validation on the transaction, stopping at the soonest failure.
+        1. Check the txid exists in memory, and is marked as valid.
+        2. Check that all scripts evaluate to TRUE.
```
dorzepowski, line 185: So the part of SPV will be running a unlocking/locking scripts?

transactions/0062.md
```diff
@@ -1,30 +1,36 @@
...
+Simplified Payment Verification formats for transmitting transactions between peers has yet to see wide adoption. This proposal advocates for complete ecosystem adoption of the principles of SPV; acknowledges that the system is secured by economic incentives, and law; and lays out a binary format for transmitting the data between parties optimizing for minimal bandwidth.
 
 Three prior formats should be mentioned:
 
-Extended Format [BRC-30](./0030.md) incorporates the utxo script for script evaluation and satoshis for checking amounts. Ideal for confirmed txs but insufficient for unconfirmed.
+Extended Format [BRC-30](./0030.md) incorporates the utxo script for script evaluation and satoshis for checking amounts. This format would still work when sending to nodes as they have a utxo lookup which is indexed by a hash of the extended data, meaning that invalid ef data would be detected immediately.
```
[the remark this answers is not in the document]
dorzepowski: Hmmm... @sirdeggen should we still refer to EF? It doesn't look like EF anymore.

transactions/0062.md
sirdeggen: Yes

arkadiuszos4chain (approved):

PR #64, opened 2023-11-20
BRC-77

dorzepowski:
### This pull request:

### Summary

TBD

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ BRC | Standard
 81   | [Private Overlays with P2PKH Transactions](./overlays/0081.md)
 82   | [Defining a Scalable IPv6 Multicast Protocol for Blockchain Transaction Broadcast and Update Delivery](./peer-to-peer/0082.md)
 83   | [Scalable Transaction Processing in the BSV Network](./transactions/0083.md)
+84   | [Proven Identity Key Exchange (PIKE)](./peer-to-peer/0084.md)
 
 ## License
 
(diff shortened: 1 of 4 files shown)


peer-to-peer/0077.md
```diff
@@ -0,0 +1,236 @@
...
+    activate Sender BUX
+    activate Sender Wallet
+    Sender BUX->>Sender BUX: Prepare transaction using dedicated keys (using dPk[i] for proper output)
+    Sender BUX->Sender BUX: Paymail host discovery
+    Sender BUX->Receiver BUX: Paymail capability discovery
+    activate Receiver BUX
+    Sender BUX->Receiver BUX: Get endpoint from BEEF tx capability
+    Sender BUX->>Receiver BUX: Send BEEF
+    Note over Sender BUX,Receiver BUX: reference generated by wallet<br/> should be used in a request
+    Receiver BUX->>Receiver BUX: Make SPV
```
sirdeggen, line 194: This whole thing is perfect except this one line.

If we run SPV here we know that the transactions is valid in the broad sense, but we cannot know which outputs belong to the recipient, and therefore do not know the payment amount.

Just noting here for further discussion.

sirdeggen (changes requested):
Recipient cannot detect which output belongs to them.

peer-to-peer/0077.md
sirdeggen: Let's just publish this and come back to fix anything which needs to be updated?

sirdeggen (approved):
Probably need to update the BRC number

sirdeggen (approved):
updated

PR #76, opened 2024-10-22
Brc96

tonesnotes:
### This pull request:

Proposes V2 of the BEEF standard with an extension to handle "known" data in BEEFs shared between parties.

### Summary

Please review and indicate likelihood for adoption.

If adopted, the Beef and BeefTx classes submitted to the SDK today will be updated to use the serialization proposed here.


tonesnotes:
Clarification, the Beef class will always support both V1 and V2 formats.

If V2 functionality (merge of txid only transactions) is used, then serialization will be in V2 format;
otherwise, by default V1 format will be used.

PR #103, opened 2025-08-25
BRC-104 - Revise encoding general message preimage

dorzepowski:
### This pull request:

### Summary

Align description of general message serialization to payload with the real implementation in go and ts SDKs

### DIFF
--- a/peer-to-peer/0104.md
+++ b/peer-to-peer/0104.md
@@ -157,31 +157,48 @@ The **order** of included headers will be lexicographical to ensure matching pre
 - If the handshake or signature fails, the server **MUST** return `401 Unauthorized` (or another 4xx suitable code) with a JSON body describing the error.  
 - If the user sets invalid data, the server **MAY** respond with `400 Bad Request`.  
 
-### 6.7 Body Encoding Rules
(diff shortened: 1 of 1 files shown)


sirdeggen (approved):
@ty-everett please take a look

dorzepowski:
> @ty-everett please take a look. My understanding is that browsers will always set default path to "/"

I test it with browser and node js version of AuthFetch, and in case of empty path, client is always setting "/" and serializing as path (so it's never set to -1).
In Go, server is setting request path to "/" when it is empty.
The only problem is with Go client, which can have empty path as empty string. This way Go client for empty path is generating different payload then the typescript client and go server.
In other words Go client is compatible with description in BRC - but it's not compatible with ts-sdk AuthFetch and Go Middleware.
That's why I would like to adjust BRC to be aligned to ts-sdk AuthFetch and Go Middleware way of treating empty path, and adjust the Go AuthFetch.

(Beside that, I made a changes in description of serialising other parts of request to align description with the implementation in ts-sdk AuthFetch)

ty-everett (approved):

PR #130, opened 2026-04-17
BRC-122: Basket Identifier Namespace Framework

sgbett:
## Summary

Proposes a backwards-compatible extension to BRC-100's basket naming rules that introduces a **structured zone** (colon-delimited) alongside the existing **flat zone**.

- Unifies existing ad-hoc prefix reservations (`admin`, `p `, `default`, `balance `) under a single grammar
- Introduces tiered namespaces: system-reserved, module-reserved, protocol-reserved (`pool:`, `token:`), application-scoped (`example.com:`), and simple names
- The colon character is not valid in any existing BRC-100 basket name, so the extension is purely additive — zero backwards compatibility risk
- Non-implementing wallets safely reject structured names at BRC-100's character validation step
- Draws structural inspiration from RFC 3986 (URI Generic Syntax) and follows the precedent of BRC-87 for naming conventions in the BSV ecosystem

## Motivation

Each new wallet-related BRC independently grabs a basket prefix with no coordination or registry. This proposal provides a framework so that future protocols can claim namespaces without collision, and wallets can validate structured names against registered prefixes.

## Files

- `wallet/0122.md` — BRC-122 specification
- `README.md` — index entry added

## Cross-references

Extends: BRC-100 (basket naming rules)
References: BRC-44, BRC-46, BRC-87, BRC-99, BRC-112, BRC-116, RFC 3986

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ BRC | Standard
 120  | [x402 Stateless Settlement-Gated HTTP Protocol](./payments/0120.md)
 121  | [Simple 402 Payments](./payments/0121.md)
 122  | [Auditable Real-time Inference Architecture (ARIA)](./apps/0122.md)
-123  | [Basket Identifier Namespace Framework](./wallet/0123.md)
+123  | [Basket Permission Scheme Registry and Governance](./wallet/0123.md)
 124  | [Multicast Transaction Frame Format](./transactions/0124.md)
 125  | [PeerPay URI Scheme for BRC-29 Payments](./payments/0125.md)
 
--- a/wallet/0123.md
+++ b/wallet/0123.md
@@ -1,217 +1,146 @@
-# BRC-123: Basket Identifier Namespace Framework
+# BRC-123: Basket Permission Scheme Registry and Governance
 
 Simon Bettison (simon@bettison.org)
 
 ## Abstract
 
-We define a hierarchical namespace framework for [BRC-46](./0046.md) output basket identifiers. Existing standards reserve basket prefixes in an ad-hoc manner ([BRC-44](../key-derivation/0044.md) reserves `admin`, [BRC-99](./0099.md) reserves `p `, [BRC-112](./0112.md) claims `balance ` as a virtual prefix). This specification unifies these reservations under a single grammar, establishes a mechanism for claiming new namespaces, and provides conventions that prevent collisions between protocols, applications, and wallet-internal operations.
+We define a governance framework for [BRC-99](./0099.md) basket permission schemes. [BRC-99](./0099.md) reserves basket identifiers beginning with `p ` for future permission schemes and establishes the format `p <scheme-id> <basket-name>`, but defines no process for registering scheme IDs, no taxonomy for classifying schemes, and no consolidated reference for the basket name reservations scattered across multiple wallet BRCs.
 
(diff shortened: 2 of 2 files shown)


ty-everett:
While BRC-87 is specific to overlays, the convention for this is already specifically handled by BRC-99. Because BRC-99 already defines a schema-based basket naming system, it would be cleaner to utilize that rather than inventing a new mechanism. If wallets needed to be concerned with both BRC-99 aware schema and name parsing and the rules defined within this new specification, it would lead to more long-term complexity and technical debt. It would be better to define any new basket-name related conventions as specific `p <scheme>` based prefixes. For example, app-scoped baskets could be defined as `p app-scoped <app originator> <name>`, where `<name>` is the name of an originator-scoped basket. Maybe we can re-purpose and re-frame this spec as a way to define and achieve that?

sgbett:
I've had a long think about this, and I've looked more closely at BRC-99 now. 

My first reading of 99 was that it was centred around permissions, and my concern was that forcing namespace/registry/protocol through p would be assigning these buckets permission based semantics that they might not need. This may not be a concern though. You make valid points about `p <scheme>`  prefixes, they could very well work just fine.

I think adding complexity is nuanced. In the short term it adds more, but I think in the long term it prevents it, and in fact aligns it with a pre-existing standard. I think the root of the problem is that the p prefix notation is the new invention, but that's probably not going to win me many friends :)

In any case, I'm less bothered about the specifics of the formatting, and more interested in the 'regsitry' aspect of it, and I think that's where you were headed too towards the end of your comment? 

How do people claim/reserve buckets, who manages this, how do collisions get handled. All tiresome stuff that feels like a lot of nonsense, but inevitably arrives whenever there is some "shared finite resource" I can already smell the squatters...

Do you want me to try re-working the proposal around the `p <scheme>` prefix? or, we can ditch the namespacing side of it and re purpose ardour the "governance" side of it? 

Perhaps create a table, pre-populate with the already reserved buckets and then moving forwards have people submit PRs  — a kind of Bitcoin Name System (the BNS, to the internet's DNS!)

Or, if there is nothing worth saving here feel free to delete it...

ty-everett:
Governance of the BRC99 namespace resources is a weak area and that would fill a compelling gap!

We currently have Registrant, RegistryClient, and BasketMap code out there. `registry-services` is probably the home for this new registry, where an operator can assign human-readable names and metadata to each of the schema registrations. ProtoMap does this for BRC43 protocols, but it does n't extend to BRC98 schemas. Nor does BasketMap extend to BRC99 schemas.

A unified, schema registration and governance specification spanning both BRC98 and BRC99 leveraging and extending `registry-services` and defining governance / conventions would allow a more formalized process to emerge, by which basket schemas and protocol schemas can be reserved for specific purposes, contact information for responsible developers, and conformance documentation published.

That way we all have a place to go when we want to register or find one of these.

sgbett:
Ah, i hadn't seen your comment before i set about re-working the original doc :) very speedy response!

I think what you are suggesting goes much further, much better than what I have proposed here. I will need to understand registry-services/ProtoMap/BasketMap first...

sirdeggen:
Merging, because these BRCs are all part of the creative process towards the ultimate goal of solid standards, feel free to update this document further in a new PR or build upon it in another BRC.

PR #131, opened 2026-04-23
Refresh BRC specs and documentation across wallets, payments, and transactions

ty-everett:
## Summary

This PR is a broad documentation/specification refresh for the BRC repository. The main goal is to reduce confusion between older historical BRC text and the behavior that implementers actually need for present BSV wallet, overlay, certificate, outpoint, payment, and storage interoperability.

The unusual part of this PR is that several existing specs are rewritten or substantially replaced rather than supplemented by new BRC numbers. That is intentional: in the affected areas, keeping obsolete wire formats and vague historical descriptions as the main text would make the repository harder to use. The updated documents preserve historical context where needed, but move the normative or guidance text toward the behavior used by the BRC-100 wallet interface, BEEF/Atomic BEEF transaction transport, Wallet Toolbox storage/sync models, and the SDK overlay/KV/certificate primitives.

## Why this is necessary

Many older BRCs were written around pre-BRC-100 assumptions, including BRC-8 transaction envelopes, mAPI-era response structures, Authrite-specific transport assumptions, older wallet route names, DPP/Paymail flows, and broad placeholders such as generic `spendingInstructions`. Since then, the repository has accumulated newer specifications around BRC-62/BRC-95 BEEF, BRC-100 wallet APIs, BRC-103/BRC-104 peer authentication, BRC-105/BRC-118/BRC-121 HTTP payments, BRC-116 permissions, and Wallet Toolbox storage.

Without this refresh, implementers have to infer which specs are historical, which specs remain normative, and which old fields have been superseded. This PR makes that orientation explicit. It also fills previously placeholder or unused BRC numbers where implementations already existed but a stable BRC document did not.

## Repository-level navigation changes

- Updates the root `README.md` and `SUMMARY.md` so BRC-35, BRC-38, BRC-39, and BRC-40 are linked as real specs rather than unused or placeholder entries.
- Adds an "Interpreting the Repository" section explaining that the repo contains both active interoperability specifications and historical design context.
- Updates category README files so implementers can see the intended current orientation for wallets, payments, overlays, outpoints, scripts, and apps.
- Fixes small index issues such as the `EXAMPLE.md` link, the `apps/README.md` typo, the scripts BRC-20 pointer, and the Paymail BEEF title.

### BRC-36: Format for Bitcoin Outpoints

BRC-36 is rewritten from a broad UTXO storage object based on BRC-8 envelopes, raw transactions, mAPI responses, and proof fields into a narrower portable outpoint object:

- `outpoint`
- `satoshis`
[...]

PR #133, opened 2026-04-24
BRC-124: Multicast Transaction Frame Format

jefflightweb:
### This pull request:

## BRC-124: Multicast Transaction Frame Format

Introduces BRC-124, a new wire format for transporting BSV transactions over IPv6 multicast and TCP/UDP unicast. This specification complements the legacy BRC-12 frame format with a standardized 92-byte header optimized for high-throughput transaction distribution, and reliable NACK-based retransmission across a scaled multicast network fabric.


sirdeggen:
Acknowledged, this will be merged after the large refactor goes in https://github.com/bitcoin-sv/BRCs/pull/131

jefflightweb:
This work is evolving slightly following testing of the retransmission system. More updates to come soon, and in the related discussion: https://github.com/bitcoin-sv/BRCs/discussions/134

jefflightweb:
BRC-124 was committed to the master branch even though it was not finalized. These changes, therefore, are meant to update the design to the current state of implementation. In particular, the sequence mechanism has been changed from embedded monotonic counters with sender/batch attribution to a hash-chain mechanism with an improved collision domain for ultra-high throughput scalability. Within about a 60 second time window at billions of TPS, the retransmission key index held by endpoints starts to be a concern and 64 bit hash functions are necessary. XXH64 was chosen for it's hardware acceleration features and ultra-low resource costs. The hash chain function incorporates sender ID as well as shard group attribution to ensure uniqueness of the retransmission key space without introducing gap issues within the shard group flows.

jefflightweb:
The more I think about transaction flow sequencing at scale, the more I wonder if a hash chain will be too slow at high throughput. The numerical approach has the advantage of natural numbers indicating gaps, rather than an unbounded number of chain links in the face of a network partition. The Sender attribution is also the wrong approach because sequencing needs to be unique per sender, per shard, and per subtree to allow for efficient retransmission on each layer, and without introducing dependencies across them.

I am considering a hybrid solution of numerical + hash identification by NACK requester. The need to have sufficient key index collision space for the retransmission infrastructure is critical, which is what made XXH64 attractive as a 64 bit function. The domain was large enough for a reasonable time window of 60 seconds to ~10 minutes, depending on overall throughput. There is a lot to consider for the solution at scale.

jefflightweb:
Sequencing is back to a hashkey identifier for the flow + sequence number. BRC-126 has been updated to support.

Further work on a more extensible frame format that can handle fragmentation of the transaction payload, as well as potentially forward error correction, will be on going in a separate BRC.

PR #135, opened 2026-04-24
BRC-12 wire format details

jefflightweb:
### This pull request:

### Summary

This PR adds missing details about the wire format, or network header used to carry BRC-12 transaction payload data. It does not change the previous spec, but adds additional detail as used by node software and other applications.


sirdeggen:
Closing — wire frame format section was merged to master in commit 1134bbe (transactions/0012.md). Only diff is minor formatting (master uses fenced code block; otherwise equivalent).

PR #139, opened 2026-05-06
BRC-126: Multicast NACK-based Retransmission Protocol

jefflightweb:
### This pull request:

## BRC-126: Multicast NACK-based Retransmission Protocol

Adds a new BRC specifying the NACK-based retransmission and endpoint discovery protocol for the BSV multicast transaction distribution pipeline.

### What it defines

- Four UDP datagram formats: ADVERT (56 B), NACK (24 B), MISS (16 B), ACK (16 B)
- Tier/preference endpoint selection with an escalation state machine
- Beacon-based discovery via site/org/global multicast scopes
- Configurable retransmit modes (multicast, unicast, suppress-ack/miss)
- Flood prevention mechanisms (dedup window, jitter hold-off, exponential backoff, fill-suppression)

### Relationship to other BRCs

- Builds on BRC-124 (Multicast Transaction Frame Format) — uses the PrevSeq/CurSeq XXH64 hash chain for gap detection
- References BRC-12 for payload format

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ BRC | Standard
 123  | [Basket Identifier Namespace Framework](./wallet/0123.md)
 124  | [Multicast Transaction Frame Format](./transactions/0124.md)
 125  | [PeerPay URI Scheme for BRC-29 Payments](./payments/0125.md)
+126  | [Multicast Transaction NACK Retransmission Protocol](./transactions/0126.md)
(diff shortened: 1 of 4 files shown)


jefflightweb:
BRC-126 was updated to support HashKey+Sequence Number, as was BRC-124.

PR #162, opened 2026-06-22
Add wallet prompt liveness contract BRC

ty-everett:
## Summary
- Add BRC-141, authored by Ty Everett, defining the Wallet Permission Prompt Liveness Contract.
- State the explicit doctrine: connected permission-required wallet requests do not use wall-clock timeouts; if the request is pending, applications should be able to assume the wallet is asking the user.
- Apply the same contract to desktop, mobile, and non-interactive/server wallet implementations.
- Add BRC-141 to the root, GitBook, and wallet indexes.

## Validation
- Pulled/rebased onto current `origin/master` after BRC-113 through BRC-140 landed upstream.
- `git diff --check`
- Verified `README.md`, `SUMMARY.md`, `wallet/README.md`, and `wallet/0141.md` all reference BRC-141; upstream BRC-113 remains `tokens/0113.md`.

PR #163, opened 2026-06-27
Add BRC-141: NotaryHash — privacy-preserving signed-hash notarization with SPV-verifiable certificates

codenlighten:
## BRC-141: NotaryHash — Privacy-Preserving Signed-Hash Notarization with SPV-Verifiable Certificates

Proposes a transaction format and a self-contained certificate for anchoring a *signed hash* to the blockchain. A signer proves they signed a specific hash; the on-chain anchor fixes that proof in time. The notarizing service never sees the document and never handles a client private key.

**Highlights**
- **Privacy:** only `SHA-256(content)` + a signature over it ever leave the client.
- **Deterministic integrity root:** a fixed length-prefixed binary encoding (not `JSON.stringify`), reproducible across implementations.
- **Post-quantum first-class:** ECDSA, ML-DSA (FIPS 204), SLH-DSA (FIPS 205).
- **Trustless (SPV) verification:** certificates carry an SPV envelope so the anchor is confirmed against block headers alone (BRC-9/10/11; BRC-74/62 encodings permitted), with no dependency on the issuing service or any indexer.
- **Merkle batching:** one transaction anchors many proofs (RFC 6962 root).

Adds `apps/0141.md` and indexes it in `README.md` and `apps/README.md`. (Happy to renumber if 141 is taken.)

A reference implementation, a dependency-light standalone verifier, and published test vectors exist (txid recomputation checked against the genesis coinbase; Merkle fold checked against the real block-170 proof). The format has been demonstrated end-to-end on BSV mainnet, verifying fully offline against independently obtained block headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ BRC | Standard
 138  | [Single-Use Signed Proofs for Request Authentication](./peer-to-peer/0138.md)
 139  | [Multicast Shard Manifest Announcement Protocol](./transactions/0139.md)
 140  | [Threshold Key Sharing and Backup via Shamir's Secret Sharing Scheme](./key-derivation/0140.md)
+141  | [NotaryHash — Privacy-Preserving Signed-Hash Notarization with SPV-Verifiable Certificates](./apps/0141.md)
 
 ## License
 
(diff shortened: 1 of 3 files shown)


sirdeggen:
Merged as BRC-220 (renumbered from proposed BRC-141 due to a numbering clash with #162), commit 70280d9 on master.

PR #164, opened 2026-06-29
BRC-142: Multicast transaction bundle frame format

jefflightweb:
### This pull request:

## BRC-142: Multicast Transaction Bundle Frame Format

Adds BRC-142, a coalescing carrier format for the BRC-12 and BRC-30 transaction frame formats. A bundle packs many small transactions sharing one shard group and subtree into a single datagram — the inverse of BRC-130 (which fragments one large transaction across many datagrams).

Motivation: multicast fan-out cost is dominated by packets-per-second, not bandwidth; small-packet BSV traffic leaves the link mostly idle while saturating the packet engine. A component that already forwards individual frames can pack same-destination transactions into one datagram, cutting the packet rate on every downstream hop while carrying the same transactions. It is an operator-side optimization — senders don't participate, and a node may bundle or not without coordination.

Key design points: one bundle = one (sender, group, subtree) flow (single HashKey + SeqNum, so it reuses the BRC-126 retransmission machinery unchanged); bundle is the retransmission unit; bounded ≤ MTU and mutually exclusive with BRC-130; default edge-decoalesce keeps the consumer contract unchanged; relays re-bucket on a BRC-139 shard-bit transition.

Changes: new transactions/0142.md; index entries in README.md, SUMMARY.md, transactions/README.md. New frame version 0x08; no changes to existing BRCs.

Continues the multicast frame series (BRC-124/126–135/139). A reference prototype validated the wire format, the packet-reduction model, and the retransmission/re-bucketing rules.


sirdeggen:
Merged as BRC-221 (renumbered from proposed BRC-142), commit 7969a72 on master.

PR #165, opened 2026-07-05
BRC-124: SSM transport note; remove stale proxy binary name

jefflightweb:
### This pull request:

### Summary

BRC-124 named the ingress proxy by a binary name (`bitcoin-shard-proxy`) that does not exist in the reference implementation, and carried no transport-mode context. Removes the stale name and adds a note that the frame is transport-agnostic — the on-wire bytes are identical under ASM and SSM — pointing to BRC-129 for SSM `(S,G)` addressing/source-discovery and naming the frame's source (the emitting proxy, `senderIPv6`). Adds a BRC-129 reference.

### DIFF
--- a/transactions/0124.md
+++ b/transactions/0124.md
@@ -4,7 +4,7 @@ Jeff Harris (jeff@lightweb.net)
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


sirdeggen:
Merged as part of docs/BRC-124 fix, commit 1e377c8 on master.

PR #166, opened 2026-07-05
BRC-126: ACK repair confirmation; correct retransmit-mode flag names, beacon-scope values and ADVERT fan-out; SSM source-discovery note

jefflightweb:
### This pull request:

### Summary

Corrects the ADVERT beacon documentation and adds SSM context.

- The `-beacon-scope` flag accepts `site`, `org`, `global`, `both`, and `all` (the spec listed only `site`/`global`/`both`); `both` and `all` are equivalent.
- Scope byte `0xFF` fans out to three ADVERTs (site + org + global), not two.
- Removes the stale `bitcoin-shard-proxy` binary name from the beacon tier table.
- Adds an SSM note: beacon addresses take the `FF3x` form under SSM (inter-domain is SSM-only per RFC 8815), and because beacon groups are control-plane their source (the retry endpoint) is discovered via per-group bootstrap source lists, not from within the group. Adds a BRC-129 reference.

### DIFF
--- a/transactions/0126.md
+++ b/transactions/0126.md
@@ -282,7 +282,7 @@ transaction source, and assigned a preference within each tier.
 
 | Tier   | Meaning                                            |
 | ------ | -------------------------------------------------- |
-| `0`    | Same AS as `bitcoin-shard-proxy` (source-adjacent) |
+| `0`    | Same AS as the ingress proxy (source-adjacent)     |
 | `1`    | One AS boundary from source                        |
 | `N`    | N AS hops from source                              |
 | `0xFF` | Static seed (no beacon received; lowest priority)  |
@@ -339,8 +339,9 @@ listener advances to the next position.
 ### Beacon Scopes
 
 Three scope bytes are defined for the ADVERT wire format. The reference
-implementation's `-beacon-scope` flag accepts `site`, `global`, or `both` (sends
-to site + global simultaneously):
+implementation's `-beacon-scope` flag accepts `site`, `org`, `global`, `both`,
+or `all` (`both` and `all` are equivalent: the ADVERT is sent to the site, org,
+and global beacon groups simultaneously):
 
 | Scope           | Scope byte | Beacon group   | Use case                                            |
 | --------------- | ---------- | -------------- | --------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


jefflightweb:
Expanded this PR beyond the original beacon-scope/fan-out corrections with two related fixes. Flagging the scope change explicitly so it is easy to review or ask me to split.

**1. ACK repair confirmation (normative).** The ACK section said the listener "cancels the gap entry immediately" on receipt, unconditionally. That is unsafe for the unicast case: an ACK asserts a retransmit was *dispatched*, not that it arrived, and the retransmit can be lost on the very path that dropped the original frame. A conformant implementation following the old text closes the gap on an unconfirmed repair, so the loss is never retried and never reported — it is counted as recovered.

The added `Repair confirmation` paragraph makes it flag-conditional: on an ACK carrying `0x02` (unicast sent) the listener MUST NOT cancel, MUST await the frame itself, and MUST resume escalation if it does not arrive. An ACK without `0x02` has no confirmation channel, so cancelling on it is still permitted (MAY), with the consequence stated. The escalation state machine and its diagram are updated to match, since they carried the same unconditional ACK edge.

**2. Retransmit-mode flag names were wrong.** The Configurable Retransmit Modes table published `-retransmit-multicast` / `-retransmit-unicast`. Neither flag exists in the reference implementation; the real names are `-beacon-flags-multicast` (default `true`) and `-beacon-flags-unicast` (default `false`), and the `beacon-flags-` prefix is meaningful because the same switch sets the corresponding ADVERT capability bit. As published, the deployment-profile examples were uncopyable. Defaults and effects are unchanged; `-suppress-miss` / `-suppress-ack` in the same table were verified correct and left alone.

I also broadened the Edge/unicast profile rationale: it was scoped to "listeners not on the multicast fabric", but unicast is also what makes a repair verifiable per (1), and a multicast retransmit is delivered to every listener in the group including those that observed no loss.

Deliberately left out: guidance on how an operator should *assign* Tier. There is a real hazard there — the spec gives Tier no default, so an implementation defaulting it to 0 has every endpoint advertising itself as source-adjacent, and equal `(Tier, Preference)` pairs leave the escalation order unspecified — but that is a separate concern from this PR and I would rather file it on its own.

sirdeggen:
Merged as part of docs/BRC-126 fix, commit 3081c0d on master.

PR #167, opened 2026-07-05
BRC-127: SSM addressing and source-discovery note

jefflightweb:
### This pull request:

### Summary

BRC-127 described its control-plane announce group with ASM-only addresses and a bare "join the group" model. Adds a note that under SSM addresses take the `FF3x` form (inter-domain SSM-only per RFC 8815) and that listeners `(S,G)`-join the announcing block-assembler source via BRC-129 bootstrap source lists — the same source addresses already used by the spec's sender include-filter. Adds a BRC-129 reference. No wire-format change.


sirdeggen:
Merged as part of docs/BRC-127 fix, commit 3fc34b2 on master.

PR #168, opened 2026-07-05
BRC-128: SSM transport note

jefflightweb:
### This pull request:

### Summary

BRC-128 (the Extended Format payload variant) carried no transport-mode note. Adds a Compatibility bullet stating BRC-128 frames are byte-identical on the wire under ASM and SSM, inheriting addressing and the SSM `(S,G)`/source-discovery model from BRC-124 and BRC-129, with the source being the emitting proxy (`senderIPv6`). Adds a BRC-129 reference.


sirdeggen:
Merged as part of docs/BRC-128 fix, commit 1430ac5 on master.

PR #169, opened 2026-07-05
BRC-130: remove 0x06 from OrigFrameVer values; SSM fragmentation note

jefflightweb:
### This pull request:

### Summary

Two corrections.

- The `OrigFrameVer` value list included `0x06` (BRC-134), but BRC-130 fragmentation is not defined for BRC-134 and the reference reassembler dispatches only `0x04`/`0x05` (with `0x00`/`0x02` as the BRC-124 default) — an `0x06` fragment would be reassembled as a plain BRC-124 frame. Removes `0x06`.
- Adds a note that fragmentation is transport-agnostic and every fragment of a transaction is emitted from the same proxy source (`senderIPv6`), so a receiver joins a single `(S,G)`; points to BRC-129. Adds a BRC-129 reference.

### DIFF
--- a/transactions/0130.md
+++ b/transactions/0130.md
@@ -209,9 +209,18 @@ For jumbo frames (pathMTU = 9000): `fragDataSize = 8848 bytes`.
 
 A listener reassembles a fragmented transaction as follows:
 
-1. **Slot allocation.** On receipt of the first fragment for a TxID, allocate a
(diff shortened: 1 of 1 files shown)


sirdeggen:
Merged as part of docs/BRC-130 fix, commit a88b2bb on master.

PR #170, opened 2026-07-05
BRC-131: SSM inter-domain addressing note for block broadcast

jefflightweb:
### This pull request:

### Summary

BRC-131 states block announcements "cross site boundaries" but gave only the ASM address `FF0E::B:FFFE`, which RFC 8815 forbids inter-domain. Adds a note that inter-domain delivery uses the SSM form `FF3E::B:FFFE` (per BRC-129), names the source receivers `(S,G)`-join (the emitting proxy, `senderIPv6`), and confirms the frame format, HashKey, SeqNum, and NACK path are unchanged across modes.


sirdeggen:
Merged as part of docs/BRC-131 fix, commit 0970c35 on master.

PR #171, opened 2026-07-05
BRC-132: SSM addressing note for subtree data

jefflightweb:
### This pull request:

### Summary

BRC-132 gave ASM-only addresses for the subtree-data announce group. Adds a note that under SSM addresses take the `FF3x` form (inter-domain SSM-only per RFC 8815), receivers `(S,G)`-join the subtree-data source (the emitting proxy, `senderIPv6`), and the data-plane source set is distributed via the shard manifest per BRC-129; frame and NACK/fragmentation paths are unchanged.


sirdeggen:
Merged as part of docs/BRC-132 fix, commit c31e576 on master.

PR #172, opened 2026-07-05
BRC-133: SSM addressing note for coinbase

jefflightweb:
### This pull request:

### Summary

BRC-133 delivers coinbase frames on the global `GroupBlockBroadcast` group with an ASM-only address `FF0E::B:FFFE`. Adds a note that inter-domain delivery uses the SSM form `FF3E::B:FFFE` (per BRC-129), and that coinbase and block-announce frames — though separate flows — share the same egress source, so SSM receivers `(S,G)`-join `0xFFFE` once. Adds a BRC-129 reference.


sirdeggen:
Merged as part of docs/BRC-133 fix, commit 0473bc4 on master.

PR #173, opened 2026-07-05
BRC-134: SSM addressing note for anchor

jefflightweb:
### This pull request:

### Summary

BRC-134 delivers anchor frames on the global `GroupBlockBroadcast` group and stresses geographically distributed reach, but gave only the ASM address `FF0E::B:FFFE` (RFC 8815 forbids inter-domain ASM). Adds a note that inter-domain delivery uses the SSM form `FF3E::B:FFFE` (per BRC-129), and that because anchors are user-submitted their SSM source set is broader than the miner-only block/coinbase sources. Adds a BRC-129 reference.


sirdeggen:
Merged as part of docs/BRC-134 fix, commit d50c35c on master.

PR #174, opened 2026-07-05
BRC-135: correct emitter HashKey group index to 0xFFFA; SSM egress note

jefflightweb:
### This pull request:

### Summary

Two items.

- Corrects the emitter `HashKey` ingredient from `0x0000FFFE` (`GroupBlockBroadcast`) to `0x0000FFFA` (`GroupBlockHeader`) — the egress channel the frame is actually sent on, matching the rest of the spec (the Multicast Group section) and the reference implementation.
- Adds an SSM note: the egress address takes the `FF3x` form under SSM, the source is the emitter (`emitterIPv6`), and the multi-emitter reliability pattern must use distinct per-emitter sources or VRRP active-standby — anycast/shared-source emission is not supported under SSM (per BRC-129).

### DIFF
--- a/transactions/0135.md
+++ b/transactions/0135.md
@@ -65,6 +65,13 @@ The egress group-id is set independently of the fabric group-id. This ensures
 BRC-135 frames reach only downstream consumers, not peer fabric subscribers or
 retry endpoints on the ingress fabric.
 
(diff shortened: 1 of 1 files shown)


sirdeggen:
Merged as part of docs/BRC-135 fix, commit 2bd3191 on master.

PR #175, opened 2026-07-07
BRC-143: Subtree data frame format

jefflightweb:
### This pull request:

## BRC-143: Subtree Data Frame Format

Adds BRC-143, the push wire form of a subtree — an ordered list of transaction node hashes plus the merkle root that identifies it — for delivery and ingest over a byte stream. A subtree names its members by hash; a receiver reconstructs the subtree and verifies it against the carried root. All frame integers are fixed-width big-endian; the sole delimiter is the node count.

Motivation: in an announce/pull system a receiver fetches a subtree by its root hash (the root is the request key, carried out-of-band in the request path) and the bytes transferred are a bare list of 32-byte node hashes. A push delivery has no request path — the subtree arrives unsolicited — so the identifying root must travel in-band. BRC-143 is that frame: root + ordered node hashes, and nothing a receiver recomputes locally. Per-node fee/size, the aggregate totals, and the conflict set are omitted (fee/size come from the receiver's transaction-meta store; the rest aren't consumed on the ingest path) to avoid inflating an already large object.

Changes: new transactions/0143.md; index entries in README.md, SUMMARY.md, transactions/README.md. No changes to existing BRCs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ BRC | Standard
 138  | [Single-Use Signed Proofs for Request Authentication](./peer-to-peer/0138.md)
 139  | [Multicast Shard Manifest Announcement Protocol](./transactions/0139.md)
 140  | [Threshold Key Sharing and Backup via Shamir's Secret Sharing Scheme](./key-derivation/0140.md)
+143  | [Subtree Data Frame Format](./transactions/0143.md)
 
 ## License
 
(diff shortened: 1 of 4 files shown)


sirdeggen:
Merged as BRC-222 (renumbered from proposed BRC-143), commit 860de7e on master.

PR #176, opened 2026-07-07
BRC-144: Block frame format

jefflightweb:
### This pull request:

## BRC-144: Block Frame Format

Adds BRC-144, the push wire form of a block — the 80-byte header, block-level counts, the ordered list of subtree roots the block references, the full coinbase transaction, the block height, and the coinbase merkle path — for delivery and ingest over a byte stream. It carries everything a receiver needs to assemble and validate the block without a follow-up fetch. Frame integers are fixed-width big-endian; consensus-defined payloads (header, coinbase, BUMP) keep their native serialisation.

Motivation: a block commits to its transactions through a header and names its subtrees in order — the block↔subtree association that binds independently delivered subtrees (BRC-143) into a block. In an announce/pull system a node is notified then pulls the full block; a push delivery has no fetch URL, so the frame carries the complete block body inline. Its field sequence mirrors the block serialisation a node already ingests — header, counts, subtree roots, full coinbase, height, coinbase BUMP — with each variable-length integer replaced by a fixed-width big-endian field. The coinbase is carried in-band because a node rejects a loose coinbase on its transaction path and block assembly needs its bytes.

Changes: new transactions/0144.md; index entries in README.md, SUMMARY.md, transactions/README.md. No changes to existing BRCs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ BRC | Standard
 138  | [Single-Use Signed Proofs for Request Authentication](./peer-to-peer/0138.md)
 139  | [Multicast Shard Manifest Announcement Protocol](./transactions/0139.md)
 140  | [Threshold Key Sharing and Backup via Shamir's Secret Sharing Scheme](./key-derivation/0140.md)
+144  | [Block Frame Format](./transactions/0144.md)
 
 ## License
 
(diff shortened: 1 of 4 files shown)


sirdeggen:
Merged as BRC-223 (renumbered from proposed BRC-144), commit 8205398 on master.

PR #178, opened 2026-07-19
Add TKQR1: Animated-QR Air-Gap Transport for Arbitrary Payloads

RexStarBSV:
### This pull request:

### Summary

This proposes TKQR1, a text framing that carries an arbitrary byte string across an optical air gap as a sequence of QR codes and reassembles it with integrity and set-consistency checks. Each frame is one line of US-ASCII: a fixed prefix, a 64-bit truncated SHA-256 payload tag, a sequence number, a frame total, and a base64url slice of the payload. The shared tag binds every frame to one payload, so a collector rejects frames from a different or stale stream, refuses to emit until the set is complete, and re-verifies the full payload hash before returning bytes. BSV currently has no standard for animated-QR transport; the closest prior art is BC-UR on BTC, which shares no wire format. The spec is grounded in a working reference implementation, two byte-compatible ports validated against a shared vector file, and is written so a second implementer can build an interoperable version from the Specification and Test Vectors alone.

### Validation

- Single new file; no index files touched, following the pattern of merged PR #156.
- The test vectors in the document reproduce exactly against both reference ports.

### DIFF
--- a/peer-to-peer/BRC-tkqr1-animated-qr-transport.md
+++ b/peer-to-peer/BRC-tkqr1-animated-qr-transport.md
@@ -1,0 +1,222 @@
+# BRC-XXX: Animated-QR Air-Gap Transport for Arbitrary Payloads (TKQR1)
+
+RexStarBSV (294282606+RexStarBSV@users.noreply.github.com)
+
+> The BRC number above is a placeholder to be assigned on acceptance.
+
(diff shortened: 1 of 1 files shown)


sirdeggen:
Merged as BRC-225 (author left a placeholder number), commit cc0bb28 on master.

PR #181, opened 2026-07-23
BRC-148: Multicast shard domain partitioning and BEEF object plane

jefflightweb:
### This pull request:

Proposes a new standard by creating a new markdown file in the appropriate directory and requests discussion and assignment of a BRC number

## BRC-148: Multicast Shard Domain Partitioning and the BEEF Object Plane

Adds BRC-148, a backward-compatible partition of the BRC-129 16-bit shard-index space into independent object planes selected by the high nibble, and the allocation of the first new plane: a BEEF object plane (domain `0x1`) carrying BRC-62/95/96 objects verbatim for overlay synchronisation. The plane shards by overlay topic — a topic name hashes to exactly one multicast group, so the group itself is the coarse topical filter — and delivery edges filter per subscriber on elected topics and BEEF encodings. A per-domain descriptor extension to the BRC-139 shard manifest lets each plane advertise and reshard its own `shard_bits` and generation independently. The transaction plane (domain `0x0`) is byte-identical to BRC-129; a BRC-129-only participant interoperates unchanged.

Motivation: BRC-88 overlay propagation discovers every interested host and sends each a copy point-to-point over HTTPS — the step that does not scale as topics and hosts multiply. This plane gives it a multicast substrate: publish once, every subscribed host receives it, with NACK retransmission. Topic-hash addressing keeps the fabric free of per-topic state, so channel counts and retransmission bookkeeping stay bounded at millions of topics; the plane reserves an eight-fold widening runway (`shard_bits` up to 15) without renumbering.

Changes: new transactions/0148.md; index entries in README.md, SUMMARY.md, transactions/README.md. No changes to existing BRCs.

PR #183, opened 2026-07-24
BRC-149: Multicast BEEF object frame format

jefflightweb:
### This pull request:

## BRC-149: Multicast BEEF Object Frame Format

Companion to #181 (BRC-148, the BEEF object plane). BRC-148 allocates the plane and, following the repo convention that frame formats are specified apart from addressing (BRC-124 vs BRC-129; BRC-143/BRC-144 vs their carriage rules), defers the concrete wire formats to a separate BRC. BRC-149 assigns FrameVer `0x09` and fixes the three byte layouts the plane needs.

The **multicast object frame** (FrameVer `0x09`, assigned here — the next code after BRC-142's `0x08`) is a 92-byte header plus payload that reuses the BRC-124 header at identical offsets, so existing classifiers, NACK retransmission (BRC-126), and fragmentation (BRC-130) require no changes. ContentID is `SHA-256d(payload)` (the reassembly and duplicate-suppression key, never the subject TxID); TopicID is `SHA-256(topic name)`, occupying the field that carries the SubtreeID in transaction frames, and is the delivery-selectivity key; HashKey and SeqNum carry the flow semantics BRC-148 defines. The payload is the BEEF object verbatim, its leading version word (BRC-62 / BRC-95 / BRC-96) acting as an encoding-capability gate.

Two unicast records complete the plane. The **submission record** is the `0xBEEF`-tagged envelope a publisher sends to operator ingress, naming one BEEF object to one or more overlay topics; on shared ports it is distinguished from a framed datagram (network magic `0xE3E1F3E8`) and a bare transaction (BRC-12 / BRC-30) purely by leading bytes, which cannot collide. The **delivery record** is the envelope a delivery edge streams to a subscriber — TopicID, length, object verbatim. Both records exist because BEEF bytes are not self-delimiting without a full structural parse, which the fabric never performs; unlike the BRC-143/BRC-144 push lanes, BEEF lanes therefore carry an explicit length-carrying envelope.

Depends on #181 for the plane definition (the `./0148.md` links resolve once it merges); the BRC-143/BRC-144 push-format references (#175/#176) are cited as in-submission. Index-row conflicts with #181 are expected and trivial (adjacent rows in README/SUMMARY/transactions README).

Changes: new transactions/0149.md; index entries in README.md, SUMMARY.md, transactions/README.md. No changes to existing BRCs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ BRC | Standard
 138  | [Single-Use Signed Proofs for Request Authentication](./peer-to-peer/0138.md)
 139  | [Multicast Shard Manifest Announcement Protocol](./transactions/0139.md)
 140  | [Threshold Key Sharing and Backup via Shamir's Secret Sharing Scheme](./key-derivation/0140.md)
+149  | [Multicast BEEF Object Frame Format](./transactions/0149.md)
 
 ## License
 
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -72,6 +72,7 @@
 * [Multicast Anchor Transaction Frame Format](./transactions/0134.md)
 * [Multicast Block Header Format](./transactions/0135.md)
 * [Multicast Shard Manifest Announcement Protocol](./transactions/0139.md)
+* [Multicast BEEF Object Frame Format](./transactions/0149.md)
 
 ## Scripts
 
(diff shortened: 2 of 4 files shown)


jefflightweb:
Marking this draft while I revise the BEEF submission-record design. The multi-topic submission record (topicCount 1..15) fans out one ingress object into up to 15 multicast frames, which — on the open/anonymous ingress class — is an amplification vector with no cost coupling. The revision makes multi-topic fan-out an authenticated/metered capability and keeps the open path single-topic. Will mark ready for review once it lands.

sirdeggen:
Merged as BRC-229 (renumbered from proposed BRC-149), commit b9a3df1 on master.

PR #200, opened 2026-08-06
Add BRC-300–307: the 1Sat Ordinals series

shruggr:
**Numbers are placeholders.** These are filed as 300–307 to keep the eight documents cross-referencing each other cleanly while drafting. They will be renumbered before merge, per this repo's lowest-available-first allocation

## What this adds

Eight documents specifying 1Sat Ordinals and the wallet scheme built on it.

| Draft | Standard |
|-------|----------|
| 300 | 1Sat Ordinals — single-satoshi tokens and origin tracking |
| 301 | Inscription envelopes |
| 302 | 1Sat assets in BRC-100 wallets (the P1Sat scheme) |
| 303 | 1Sat collectables in BRC-100 wallets |
| 304 | 1Sat ordinal collections |
| 305 | BSV-21 fungible tokens (JSON, legacy) |
| 306 | BSV-21 fungible tokens (binary) |
| 307 | Sigma — transaction-bound script signatures |

300 and 301 specify what has been running on mainnet since 2023. 305 documents the deployed JSON encoding of BSV-21; 306 specifies the script-native binary encoding intended to replace it for new tokens. 307 is the first written specification of the Sigma signature construction, including a test vector verified against the indexer that has been validating it in production.

## Relationship to existing standards

- **BRC-147** is marked superseded by 303, which is the P1Sat-aligned successor of that basket profile.
- **BRC-150** is retargeted from basket `1sat` to `p 1sat ordinals`. The proof format is unchanged; only the basket profile it is scoped to moves. Its author is credited unchanged.
- **BRC-98/99/111** provide the `p ` prefix that 302 uses as its scheme identifier.
- **BRC-48** and **BRC-35** are cited by 302 for the signed-PushDrop template. Note that BRC-48 specifies the spend-time signature in the unlocking script and does not describe the optional signature carried as a pushed field; 302 names the field and points at BRC-35 for what it covers, but the ecosystem may want that written up properly somewhere.

## Notes for reviewers

- 302 does not fix a BRC-43 security level, and the level governs exactly one thing: key derivation. A wallet applies the standard level semantics when an application requests a public key under `p 1sat`, persisting protocol grants alongside basket grants. Spending is unaffected — the scheme presents an approval for every action that creates or spends basket outputs, whichever level a row declares, which is stricter than any standing grant. Each output records the level it was created with, so rows written at different levels remain independently spendable.
- 302 has the wallet classify an operation by parsing the output locking scripts and the templates of the spent inputs, so a single structured approval describes what the transaction actually does. A template the wallet cannot classify is described to the user as unrecognized; nothing in the scheme requires refusing it, so a novel template needs no change to the permission layer before it can be used.
- 305 states that the token id inside a BSV-21 inscription is underscore-form only. That encoding predates the dotted outpoint form and is fixed by what is already on chain.


GenericCPU:
## Proposed wallet contract (HandCash / Brandon)

Thanks for writing down the wider 1Sat surface — **300–301 and 304–307 look like the right docs win** (core, envelopes, collections, BSV-21, Sigma). This comment is only about the **wallet filing contract** in 302/303 and the BRC-150 retarget.

### Proposed law (no dual-storage model)

| Layer | Law |
|-------|-----|
| **Storage basket** | `1sat` |
| **Permissions** | `p 1sat <scope>` (`*`, `collection:…`, `creator:…`, `origin:…`) |
| **Scheme** | `1sat` under `p 1sat` |

`p ` (BRC-98/99) is for **scoped grants and prompts**, not on-disk inventory names.

**Normative behavior:**
1. File collectables in storage basket **`1sat`**.
2. Apps request plain `1sat` or `p 1sat <scope>`.
3. Wallets normalize inbound `p 1sat …` → storage `1sat` and apply scope in the grant/prompt.
4. Do **not** define `p 1sat ordinals` (or any other `p …` string) as the collectable storage basket.

Scoped app access is solved by permissions + normalize. Storage does not need a `p` spelling.

### Other concrete asks
- **BRC-150:** keep proof format; target storage class `1sat`.
- **CI:** display + spend + `provenance` in one object; spend-only must not be the sole shape.
- **`id:` tags:** optional for interop.
- **Soft-latch:** if cited, use **BRC-154** — official `wallet/0153` is already *Action References*.
- **303 co-authorship:** please confirm with me before listing; credit 147 authorship either way.
- **307:** independent verify of the Sigma test vector before merge.

### Merge shape
Keep the series. Set **302 / 303 / 150** to: permissions speak `p 1sat`, storage files `1sat`, provenance in CI when available. Chain theory docs stay as-is.

Happy to iterate on exact 302/303 wording.

PR #210, opened 2026-08-08
BRC-159/160 1Sat origin + inscriptions; 150 continuity; 158 Outpoint BEEF

shruggr:
### BRC-159 (`tokens/0159.md`) — new
1Sat origin tracking: 1-satoshi chains, origin, sat ordering, transfer, outpoint→origin walk.

### BRC-160 (`tokens/0160.md`) — new
`ord` inscription envelopes on 1-sat outputs (companion to 159).

### BRC-150 (`tokens/0150.md`)
- Ordinal continuity on path hops (input sources in BEEF).
- References 159/160 for origin model and envelopes.
- Informative walk lives in 159.

### BRC-158 (`transactions/0158.md`) — new
**Outpoint BEEF**: outpoint-subject counterpart to Atomic BEEF (`0x16a7beef` + tip outpoint + BEEF). Stand-alone wire envelope.

## Test plan
- [ ] Spec review of 159/160
- [ ] 150 still valid for existing HandCash packages (additive MUSTs on beef contents)
- [x] Confirm `0x16a7beef` unused


shruggr:
@GenericCPU please review

ORDNET:
Read the full diff — the ordinal-continuity fix is correct and needed: spend-as-input alone never proved the sat on multi-input hops. Sharing our review here as well so it lives with the PR (posted earlier in Discord). Four notes:

**1. `proveOrigin` is a verifier, not a discovery algorithm — say so.**
The walk breaks on any non-1→1 hop and returns `origin = current`. For a sat that was isolated, later consolidated, and re-isolated, that returns the *re-isolation* outpoint as origin — the true origin (and its inscription) lies further up, behind a hop the walk refuses. As verification against a *claimed* `provenance.origin` this fails safe (unproven), but as written ("return origin") it reads as discovery, and there it is wrong. Suggest stating that a non-1→1 upstream hop yields "origin not provable from this bundle", not "this is the origin".

**2. BRC-158 needs a profile identifier in the wire.**
The graph is "profile-defined" and validation is "fail if any transaction *required* is missing" — but required per *which* profile? The envelope carries no profile byte, so a consumer of `0x16a7beef` cannot know which completeness rules to apply. One varint profile id (`0x01` = 1Sat tip→origin) makes the format self-describing. Related: Atomic BEEF's "unrelated txs present → fail" was dropped without a replacement bound — a SHOULD NOT on unrequired transactions plus a size-refusal rule (BRC-150 already has one) keeps bundles verifiable before they are parsed.

**3. The inductive hop leaks outside the bundle.**
Receiver step 5 allows `T` to be obtained outside `beefB64` — but rule 7's ordinal continuity on that hop needs *T's input source transactions* too. Where do those come from? Suggest stating they must accompany `T`, or that continuity on the inductive hop is deferred.

**4. Composition offer.**
BRC-150/158 prove *lineage* — this tip descends from this origin. What lineage cannot prove is *canonicity* — that this origin is THE first-valid claim of a name (a perfect proof of descent from a second claim is still a perfect proof). That is the layer we run at ODNCA: the full name state is committed on-chain as a merkle root (genesis commit at block 961546; daily prev-linked chain), and every name has a live proof object (leaf + merkle path + commit anchor) at `https://odnca.org/proof/<name>`. A p2p payment response carrying **158 for continuity + the commitment proof for canonicity** gives the sender the complete "this destination provably holds the canonical name". Happy to align our proof object with the 158 envelope, and our frozen indexer vectors (multi-input hops included) can double as test vectors for `proveOrigin`. Ruleset, vectors, and the reference verifier: https://odnca.org/transparency.dc.html

— ORDNET / ODNCA

shruggr:
**1. `proveOrigin` / origin discovery**

On `proveOrigin`: this walks the 1-sat path to verify a claimed origin. A non-1→1 hop means this outpoint **is** the origin — further crawling is not required. It is not an incomplete proof with the real origin further up. No change needed for that behavior.

shruggr:
**2. Profile id on BRC-158**

BRC-158 is an outpoint proof bundle. It is not specific to ordinals or any one proving logic — whatever system needs to prove something about that outpoint can use it. Completeness rules stay with the application, not on the wire. No profile id.

shruggr:
**3. Inductive / settle hop**

The proof is for an outpoint in the BEEF. That can be the full tip→origin chain, or it can prove the input being spent by a new transaction. In the latter case the receiver only checks that the new transaction spends that input — it is not a partial proof of the new tip with pieces missing outside the bag. If the new tip is included in the BEEF, the chain is complete whether or not that transaction is broadcast or mined.

shruggr:
**4. Composition / canonicity**

BRC-150/158 only prove tip→origin lineage for an outpoint. Broader claims (for example that an origin is the canonical name in some registry) are a different layer and out of scope here. The bundle already carries the full transaction chain for that outpoint, so other protocols can parse what they need from it or run additional validation on top without changing this format.

ORDNET:
Thanks for going through all four properly — appreciate it.

1, 2, 3 — agreed on all three. Framing proveOrigin as a verifier rather than a discovery walk resolves my first note entirely: a non-1→1 hop being the origin for the claimed outpoint is correct, and my concern only applied if it were presented as discovery. On the profile id — you're right, and it's the better design: keeping BRC-158 a dumb outpoint-proof bundle with completeness rules living in the application is exactly the separation that keeps it reusable. And the inductive hop is clear now: source txs in the BEEF, or the receiver checks only the spend. No changes needed on my side for any of these.

4 — this is the one I want to build on, and you've framed it exactly right. "A different layer, on top, without changing this format" is precisely the boundary I was reaching for. Lineage (158) and canonicity (is this origin the first-valid claim of a name under a ruleset) are genuinely different questions, and they shouldn't be entangled in one envelope. So rather than push anything into 158, the right move is a small separate BRC that references it — canonicity-proof-by-reference, consuming a 158 bundle as its lineage input and adding the merkle-inclusion-against-an-on-chain-state-commitment on top. That's the micro-protocol model you described in Discord, and honestly it's how we built ODNCA too — smallest enforceable pieces, composed by reference.

If that's welcome, I'll draft it as a proper BRC (Abstract / Motivation / Specification / test vectors) and open it as its own PR that cites 210, so it stands or falls on its own merits without touching yours. Our state-commitment ruleset and frozen vectors are already public to build it from.

One correction I owe you, going the other way: reviewing the 1Sat spec again for this, I see body values should not be concatenated across multiple data pushes — a single push. We had multi-push handling in a resolver of ours; that's on us to bring in line with the spec, not the other way around. Good reminder that the spec is the authority.

GenericCPU:
hey guys just got back after the weekend there are a few incompatibilities here I will need to address.

shruggr:
I didn't intend for it to be merged yet, but I didn't have it as a draft apparently. My bad. @GenericCPU can you make another PR please and I'll add my feedback there. Sorry about that. My PR for 147 is still a draft.

GenericCPU:
https://github.com/bsv-blockchain/BRCs/pull/214

Please review this PR

PR #214, opened 2026-08-10
docs(BRC-150): sat-ordering MUSTs + scalability; notes on 158/159/160

GenericCPU:
## Summary

Single follow-up to #210 (supersedes #215).

### BRC-150 (normative + informative)

1. **Path edges** MUST be BRC-159 1→1 hops (not mere spends).
2. **Settle inherit** MUST require sat ordering on `T→P` + known preceding sats.
3. **`beefB64`** prefers full BEEF with preceding sources; AtomicBEEF only when that set survives.
4. Short Pass / Fail / Settle hop examples.
5. **Scalability (informative):** Prefer extend/reuse/omit/pin; Avoid full re-walks, AtomicBEEF-as-shortcut, content in remittance, unbounded hydrate.

### Other BRCs (light notes only)

* **158** — profile may include non-ancestors; size grows with profile; do not nest in `beefB64`; see 150 *Scalability*.
* **159** — sat math is cheap; proof packages / size live in 150.
* **160** — large bodies stay off remittance; use content URLs.

## Test plan

- [ ] 150 receiver steps 5 + 7 and sender path construction
- [ ] Scalability Prefer/Avoid readable without prior Discord context
- [ ] 158 notes do not re-open putting Outpoint BEEF inside `beefB64`
- [ ] 159/160 changes stay informative one-liners

### DIFF
--- a/tokens/0160.md
+++ b/tokens/0160.md
@@ -163,7 +163,7 @@ Those may require [BRC-159](./0159.md), this envelope BRC, or both.
 
 * **Content authenticity** — An envelope proves data was in a script at an outpoint; it does not by itself prove the token's current outpoint still descends from that origin (see [BRC-159](./0159.md) and [BRC-150](./0150.md)).
 * **Parent without spend** — Field 3 alone is not proof of parentage; only spending the claimed parent in the same transaction establishes the link.
-* **Large bodies** — Very large pushes affect relay and storage; application limits are out of scope here but implementers should plan for them. A [BRC-150](./0150.md) package that covers the origin transaction already carries that origin’s locking script inside `beefB64`, so the envelope body is present in the BEEF — there is no separate “inscription body” field to add. Do **not** duplicate the body outside the BEEF (extra remittance keys, parallel blobs). Oversized origins inflate the bag; wallets MAY omit remittance rather than truncate, and SHOULD serve media for display from content URLs / OrdFS when they are not unpacking the BEEF for UI.
+* **Large bodies** — Very large pushes affect relay and storage (including [BRC-150](./0150.md) bags that cover the origin tx). Application limits are out of scope here.
 * **Content selection** — Showing the wrong envelope for a given app (e.g. latest vs origin) is an application concern, not fixed by this BRC.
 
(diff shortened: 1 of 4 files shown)


shruggr:
All looks good to me and were incomplete in the last round of updates. Thanks

tokens/0160.md
```diff
@@ -163,7 +163,7 @@ Those may require [BRC-159](./0159.md), this envelope BRC, or both.
 
 * **Content authenticity** — An envelope proves data was in a script at an outpoint; it does not by itself prove the token's current outpoint still descends from that origin (see [BRC-159](./0159.md) and [BRC-150](./0150.md)).
 * **Parent without spend** — Field 3 alone is not proof of parentage; only spending the claimed parent in the same transaction establishes the link.
-* **Large bodies** — Very large pushes affect relay and storage; application limits are out of scope here but implementers should plan for them.
+* **Large bodies** — Very large pushes affect relay and storage; application limits are out of scope here but implementers should plan for them. Do not put inscription bodies into [BRC-150](./0150.md) remittance — envelope presence at origin is enough; serve media from content URLs / OrdFS.
```
shruggr: Don't you have to have the inscription bodies in BRC-150? You have to prove the inputs of that transaction, so you have the inscription in the beef.

GenericCPU: You’re right — if the origin tx is in `beefB64`, the envelope body is already there in the locking script. The old wording read like we were saying bodies aren’t in the package at all.

Clarified in 0160 / 0150: verification reads the envelope **in place** from the BEEF. The MUST NOT is only against *duplicating* the body outside `beefB64` (extra remittance keys / parallel blobs). Large inscriptions still inflate the bag itself; omit-vs-truncate and OrdFS/content URLs are about size / display, not about leaving the body out of the proof.

GenericCPU: Simplified further: dropped the “don’t put bodies in remittance” lecture. Origin scripts (and thus bodies) are already in `beefB64`; 0160 just keeps a one-line size warning.

sirdeggen (approved):

PR #215, opened 2026-08-10
docs(BRC-150): scalability notes — extend hot path, omit oversized bags

GenericCPU:
## Summary

Informative **Scalability** section for BRC-150 after #210.

Remittance that includes preceding input sources for sat ordering grows with tip depth. The bottleneck is depth × full BEEF, not BRC-159 math or BRC-160 envelopes.

**Prefer:** extend/reuse a verified remittance (O(1) per send), omit when over budget, pin local verdicts and discard giant cold-start BEEFs, attach settle-hop funding sources with `T`.

**Avoid:** full tip→origin rebuild on every send/list, AtomicBEEF / inlined Outpoint BEEF as a size shortcut that drops funding sources, embedding inscription content, unbounded genesis hydrate on the UI thread.

Complements (does not replace) normative follow-ups in #214.

## Test plan

- [ ] Section is clearly informative (no new MUSTs that conflict with #210)
- [ ] Prefer/Avoid list matches sender step 2 + Compatibility size omit
- [ ] Readers see why AtomicBEEF-as-shortcut is called out without reopening the 158-in-`beefB64` debate

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tokens/0150.md
+++ b/tokens/0150.md
@@ -113,6 +113,28 @@ On success, the receiver MAY treat `provenance.origin` as the proven origin for 
 * Partial BEEF (structurally valid but missing a path transaction or a required input source transaction) MUST fail verification.
 * [BRC-158](../transactions/0158.md) **Outpoint BEEF** is a binary wire format for the same tip→origin proof (outpoint-subject BEEF envelope). It is an alternative to this JSON remittance, not a `beefB64` encoding option inside it.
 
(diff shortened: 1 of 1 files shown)


GenericCPU:
Superseded by #214 — scalability Prefer/Avoid folded into the same BRC-150 follow-up, with light notes on 158/159/160.

PR #216, opened 2026-08-10
docs(147): content + settle tags/CI (defer id: to 164)

GenericCPU:
## Summary
- Adds `content` (and `content:` tag) for derivative / reference 1sats (shared on-chain media).
- Settles the remaining BRC-147 tag/CI clarifications that were drafted in #206: origin forms (`origin` vs `origin:<…>`), `type:`, tags vs `customInstructions`, prefer CI `name`.
- Does **not** define `id:` — wallet-local list keys belong in BRC-164 (#219). This profile only says writers MAY stamp `id:` when adopting 164, and receivers MUST ignore counterparty `id:`.

## Why one PR
Avoid a second 147 PR. #206’s 1sat-specific bits land here; `id:` stays on #219. Please close #206 as superseded for 147.

## Test plan
- [ ] Spec review: claims vs BRC-150 proof boundary still clear
- [ ] Confirm `id:` is only referenced, not redefined
- [ ] Confirm origin forms + `type:` match intended 1sat practice
- [ ] #206 closed / redirected after this merges

### DIFF
--- a/tokens/0147.md
+++ b/tokens/0147.md
@@ -19,6 +19,8 @@ It does **not** redefine 1Sat Ordinals origin theory. Offline tip→origin proof
 
 Existing deployments (including HandCash Desktop) already use basket `1sat` with overlapping tags and JSON instructions. This BRC records that convention so independent wallets can interoperate without trusting a single vendor.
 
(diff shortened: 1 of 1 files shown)


GenericCPU:
Follow-up commit on this branch: Motivation now cites the real-world indexer limit that motivated the field.

Large 1Sat collections (Pixel Foxes ~10M items) already find collection-filtered holder queries unusable in practice, while per-origin `/content/{origin}` remains workable. Without a named `content` remittance field, wallets that rebuild CI from `origin`/`name`/`provenance` drop the shared media pointer and force the receiver back onto that fragile indexer path. Forwarding `content` keeps derivative display peer-to-peer.

PR #219, opened 2026-08-11
BRC-164: Output identity tags (id:) for BRC-100 wallets

shruggr:
## Summary

Adds **BRC-164**: ordinary output tag convention `id:<key>` as a stable, queryable list key for held BRC-100 outputs.

- No BRC-100 schema changes
- No special wallet peel/inject/strip — stamp and query like any other tag
- Conforming tools/wallets that adopt the profile SHOULD stamp `id:<key>` on create/custody
- Not a global asset, token, origin, or on-chain id (those stay in basket-profile tags)

## Test plan
- [ ] Spec review (prefix, meaning, ordinary-tag behavior)
- [ ] Follow-up: 1sat-sdk (and any tools) stamp/query `id:` per this BRC

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,7 @@ BRC | Standard
 158  | [Outpoint BEEF](./transactions/0158.md)
 159  | [1Sat Ordinals — Single-Satoshi Tokens and Origin Tracking](./tokens/0159.md)
 160  | [1Sat Ordinals — Inscription Envelopes](./tokens/0160.md)
+164  | [Output Identity Tags for BRC-100 Wallets](./wallet/0164.md)
 168  | [Verifiable Time Allocation](./apps/0168.md)
 169  | [Universal Handle Addressing and Resolution for the Metanet](./peer-to-peer/0169.md)
 210  | [Derived Collectibles](./apps/0210.md)
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -48,6 +48,7 @@
 * [Wallet Permissions and Counterparty Trust](./wallet/0116.md)
 * [Basket Permission Scheme Registry and Governance](./wallet/0123.md)
 * [Action Reference Labels for BRC-100 Wallets](./wallet/0153.md)
+* [Output Identity Tags for BRC-100 Wallets](./wallet/0164.md)
 * [Pluggable Backup Services for BRC-140 Share Vaults](./wallet/0154.md)
 * [Pull-Based Receive Discovery](./wallet/0155.md)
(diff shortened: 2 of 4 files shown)


GenericCPU (commented):
Thanks for this — cleanly scoped and the “ordinary tag, no peel/inject, not an on-chain id” boundary is the right one. A few review notes from a wallet that already stamps `origin:` / `collection:` / `app:` for 1sat:

### 1. `id:` is very short to leave unreserved

The security section admits `id:` “may appear in other informal tag sets” and then declines to reserve exclusivity of the two-letter prefix. That’s the sharp edge: a wallet (or tool) that already uses `id:` informally will silently satisfy a BRC-164 `listOutputs` filter with the wrong row.

Options that would harden this without changing the BRC-100 shape:

- Reserve `id:` for this list-key role for adopters (same stance BRC-147 takes on `origin:` / `app:`), **or**
- Prefer a namespaced form in the normative text (e.g. `outid:` / `rowid:`) and keep short `id:` as a MAY alias.

Leaving it as “MAY document a longer key form” puts the collision cost on every reader.

### 2. Confirm storage-local meaning vs basket-profile tags

We already key held tips on `origin:` and now also `collection:` / `app:`. Those are cross-wallet claims about the inscription. BRC-164’s `id:<key>` is deliberately a *storage-local* handle with no cross-wallet meaning — which is fine, but worth stating even more bluntly for receivers:

- On inbound remittance / `internalizeAction`, non-adopters (and even adopters) MUST ignore a counterparty-supplied `id:` and stamp their own if they adopt the profile.
- Basket profiles that need asset/origin identity keep using their own tags; `id:` MUST NOT be treated as a substitute for `origin:` / `bsv21:…`.

If that’s the intent, a one-liner under Behavior / Out of scope would close it.

### 3. README / SUMMARY ordering nit

In both `wallet/README.md` and `SUMMARY.md`, 164 is inserted **before** 154 / 155, breaking numeric order. Root `README.md` is fine. Minor, but easy to fix in a follow-up.

Low risk for non-adopters either way — opt-in, and nothing here forces a stamp. Happy to stamp once the prefix collision story is settled.

GenericCPU:
Proposed rename against this branch: `id:` → `outid:` — https://github.com/shruggr/BRCs/pull/$(gh pr list --repo shruggr/BRCs --head GenericCPU:docs/brc-164-outid --json number -q '.[0].number')

GenericCPU:
Proposed rename against this branch: \`id:\` → \`outid:\`

https://github.com/shruggr/BRCs/pull/1

PR #224, opened 2026-08-16
Add BRC-300: Sponsored Onboarding and Wallet Choice (supersedes #223)

vincemedia:
Supersedes #223, which I am closing in favour of this.

## Thank you

**@deggen** raised this in review, and it was the best objection the document has had:

> A serious security attack vector in this proposal is that I can buy a prominent spot to gather a userbase to my wallet software which may be malware of some sort which waits for weeks before rugging the users all at once.

It was correct. Checking it against the text, the entire defence turned out to be one clause buried in the Security Considerations, saying choosers "MUST resolve install targets against the catalog" and "SHOULD pin vendor identity keys". Meanwhile section 4.4's offer carried the full wallet object, so a sponsor supplied its own name, icon, description and install target; section 4.3.8's eligibility list was purely economic and never looked at the wallet; and section 8.3.3 let a house exclude a *site* with no equivalent for a *sponsor*.

Two things make it worse than the objection stated, and both are now in the document.

**The ranking rule selects for the attacker.** An honest vendor's bounty is bounded above by what a customer is worth to it. A vendor intending to steal is bounded by what it expects to steal, which scales with everything those users will ever hold. So it outbids honest competitors at every auction, and section 4.3.1 forbade the house from considering anything else. **The quantity this document ranks on correlates with intent to defraud.** No sorting rule fixes that, because the problem is who gets into the set being sorted.

**The retention mechanism certified the attack.** Section 7.5 pays a site when the funded key is still transacting ~30 days later. A patient attacker's victims satisfy that perfectly, so the machinery built to prove quality paid sites for delivering them and stamped the settled record `retained`.

## What changed

The governing principle, now stated outright in 4.3.8: **the auction re-orders an already-vetted set and never admits to it. Money buys position among vetted wallets. It never buys entry.**

1. **Catalog listing is a precondition of bidding** (4.3.8 condition 10). A campaign's wallet must already be in the catalog serving that site's unsponsored slots, and the catalog entry must agree with the campaign on vendor, name and platform support. The sponsored and unsponsored halves of a chooser now clear the same bar.
2. **The offer names a wallet; it cannot describe one** (3.2.1, 4.4.2). The `wallet` object is gone from the offer, replaced by `vendorKey` and `walletName`. Every card field is read from the catalog entry. There is no longer any field in which a sponsor could ship its own icon, description or install target, which is the same move as the closed schema of 4.4.9 applied to content rather than styling.
3. **The sponsor key must be attested at the vendor's own domain** per BRC-68 (4.3.8 condition 11). An anonymous sponsor cannot bid.
4. **A house can stop serving a sponsor, and a catalog can delist the wallet** (8.3.5), symmetric with the existing site rule. Delisting is the effective remedy, because condition 10 means a wallet out of the catalog cannot be bid for anywhere.
5. **Retention is not a safety signal** (7.5.5). No party may present a `retained` outcome or a paid share as evidence a wallet is safe, and a catalog may not use retention as an admission or ranking criterion.
6. **Sponsorship is not endorsement** (3.5.5). The disclosure expansion must now say the site is not vouching for the wallet.
7. **The adverse selection is named**, in the Security Considerations, in 8.5 as a third thing the design does not prevent, and in the ledger, which gains a row reading "the wallet in a sponsored slot was vetted by somebody — **Asserted**".

## What this still does not fix

Stated plainly in the document rather than left for the next reviewer.

A wallet vetted honestly and turned malicious a year later passes every check here. The remedy is delisting, and **catalog governance is not specified by this document** and only sketched in BRC-137 section 3. Section 9 now lists it as unspecified and explains that it is load-bearing rather than a tidy-up: who may list a wallet, on what evidence, who may delist and how fast, and how an exclusion propagates between catalogs. The security of the sponsored half of every chooser rests on a standard nobody has written yet, and that is probably the strongest argument for writing one.

The bounty also makes the attack cheaper to run, since the attacker pays its victims to install with money it expects to recover from them. That is inherent to paying people to onboard and is not solvable inside this design.

## Verification

98 checks to 100, still no third-party imports and no network.

```
```

The two new checks assert that the offer's canonical form carries no card content and does name a wallet instead. Appendix A.4 was regenerated and now reads the absence off the canonical form directly.

## Re-review

@deggen would you take another look? Points 1, 2 and 4 are the ones that answer your objection directly, and I would particularly value your view on whether making catalog listing a hard precondition is the right containment or whether it just relocates the trust problem into a document that does not exist yet. I think it relocates it honestly, but you found the last thing I was wrong about.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,7 @@ BRC | Standard
 226  | [Miner-Enforced Resale-Royalty Covenant Tokens (OP_PUSH_TX)](./tokens/0226.md)
 227  | [Frictionless On-Chain Onboarding via Pre-Funded Claimable Tokens](./apps/0227.md)
 228  | [Unlinkable Payments under the Identity Paradigm](./payments/0228.md)
+300  | [Sponsored Onboarding and Wallet Choice](./apps/0300.md)
 
 ## License
 
--- a/apps/0300.md
+++ b/apps/0300.md
@@ -1,0 +1,1182 @@
+# BRC-300: Sponsored Onboarding and Wallet Choice
+
+Crumbs, Siggi Oskarsson, Luke Rohenaz
+
+## Abstract
+
+This document specifies what happens when a person with no wallet arrives at a Metanet application and is invited to connect one.
+
+It defines three things. The **connect surface**: the control a site offers, the vocabulary it uses, why that vocabulary is not interchangeable with the vocabulary of signing in, and the manifest entry by which a site declares whether an auction runs in front of its users at all. The **chooser**: the dialogue that opens when no wallet is reachable, which MUST present at least two wallets from at least two vendors, because a list of one is a product recommendation and a list of two is a demonstration that the account is not the site's to keep. And the **onboarding auction**: sealed-bid, first-price, decided in under 250 milliseconds, in which wallet vendors bid for placement by committing to fund the newcomer's wallet.
+
+The bid is the bounty. A vendor wins on the amount the person receives and on nothing else; what it separately pays the auction house is a billing matter that MUST NOT influence which campaign wins. The amount is never transmitted, only a commitment to it, so a client that is never given a number cannot rank by it and a scraper that reads every field of every offer learns nothing it could sort. Every bid is published and only the winner's is opened, which lets any losing sponsor establish from its own key that its bid was in the auction and that the winner outbid it.
+
+Targeting is a **client-evaluated predicate**: a campaign declares its audience, the chooser checks it locally against [Bitcoin Attestation Protocol](https://github.com/icellan/bap) attestations or [BRC-52](../peer-to-peer/0052.md) certificates the person already holds, and discards what does not match. Nothing goes back. A bounty MUST be locked before it is bid, as a [BRC-227](./0227.md) voucher batch derived from the sponsor's own key, and the commitments come from the same key and the same reference, so commitment `i` and voucher `i` are the same slot and the sponsor stores nothing.
+
+## Motivation
+
+[BRC-137](../opinions/0137.md) established that an application meeting a wallet-less visitor MUST NOT show them an error, MUST detect their operating system, and MUST resolve install recommendations from a catalog. That solved the dead end. It left three things open, and each is where the Metanet's central claim is either made or quietly dropped.
+
+**One wallet is not a choice, and a choice is the whole argument.** What a Metanet application offers a newcomer is not better sign-in; it is that the account is theirs and travels. That claim is unfalsifiable in the abstract and immediate in a list: two wallets, from two companies, either of which opens the same identity. BRC-137's catalog can return one entry and conform. This document requires two, from two vendors.
+
+**Nobody is paying for the funnel, so the funnel does not exist.** A vendor's most expensive problem is a person who has heard of self-custody and will not install anything to try it. A site's most expensive problem is a visitor who bounces at the wallet requirement. The two are complementary and unconnected. Advertising solved the analogous problem with a real-time auction and then spent twenty years demonstrating every way that mechanism can be turned against the person it is shown to. The mechanism is not the problem. The direction of the payment is.
+
+**A bounty that can be seen is a bounty that will be farmed.** If the offer carries the amount, the amount reaches the device, and whoever wants to harvest it reads it out of the network tab, sorts, installs the top one, and repeats. Asking an interface not to render a number it possesses is not a control. So the offer carries a commitment instead, which the vendor cannot later reduce and the recipient cannot read.
+
+### What is actually being specified
+
+Wallet onboarding is the occasion for this document rather than the whole of its subject, and a reader who takes it for a standard about install prompts will misjudge which rules matter.
+
+What is specified is **an attention market whose ranking variable is the benefit to the person whose attention is bought, made structural rather than promised.** Three inversions produce it, each enforced by what the wire carries:
+
+1. **The payment points at the person.** A slot is won with the amount delivered to the newcomer, and section 4.3 forbids ranking on anything else. What the buyer pays the intermediary is a separate number that no ranking may consult.
+2. **No profile is assembled, because the description travels outward.** The audience is declared by the buyer and evaluated on the person's device. Section 4.2 forbids an identifier in the request, and at the moment it is issued there is not one in existence to forbid.
+3. **The sortable number is never transmitted.** Only a commitment to it. A client not given a number cannot rank by it, so the property survives a careless implementation and a hostile one alike.
+
+Nothing in the three is specific to wallets. What is specific to wallets is section 3, and section 3 is there because a market that recommends one product to everybody is not a market.
+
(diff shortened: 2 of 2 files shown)


vincemedia:
@deggen re-review request, and thank you.

Your objection was right and it was the most useful thing anybody has said about this document. The defence I thought was there turned out to be one clause in the Security Considerations, while section 4.4's offer happily carried a sponsor-supplied name, icon, description and install URL, and section 4.3.8's eligibility list never looked at the wallet at all.

Three changes answer you directly:

- **4.3.8 conditions 10 and 11** make catalog listing and a BRC-68 domain-attested sponsor key preconditions of bidding. Money buys position among vetted wallets, never entry.
- **3.2.1 and 4.4.2** strip card content out of the offer entirely. It now carries `vendorKey` and `walletName` and nothing else about the wallet, and every field on the card is read from the catalog entry, so there is no field left in which a sponsor could ship its own icon or install target.
- **8.3.5** lets a house stop serving a sponsor and a catalog delist the wallet, which the previous revision allowed for sites and not for vendors.

Two things your objection implied that I have also written down. The ranking rule *selects* for the attacker, because a thief's willingness to bid is bounded by what it expects to steal while an honest vendor's is bounded by customer value, so the quantity the auction sorts on correlates with intent to defraud. And section 7.5's retention share would have paid sites for delivering victims, since a patient attacker's users satisfy the retention condition perfectly; 7.5.5 now forbids treating retention as a safety signal.

The question I would most value your view on: making catalog listing a hard precondition contains the attack, but it relocates the trust into catalog governance, which this document does not specify and BRC-137 only sketches. Section 9 item 3 now says so explicitly. Is that honest containment, or have I just moved the hole somewhere nobody is looking?

sirdeggen:
It's a good question. Let's discuss at length because there's been talk in the past about hiring a company to audit wallet providers and then certifying them as the BSVA as having met BRC-100 and IT sec standards. There's also potential to limit things to ISO certified IT sec standards companies or something like that. It does move the problem elsewhere, but that was the point I think. The design of an anti manipulation / credibility system is likely a whole BRC of its own.

sirdeggen:
merge blocked by the substantial conformance vector code in the PR. Ideally BRCs should remain documentation / idea explainers rather than code to run, that should move to its own repo and we can link to it from the BRC document please.

vincemedia:
@sirdeggen done, and thank you: the PR is now two files, `apps/0300.md` and the standards-table row. Nothing executable.

The vectors have moved to their own repository at **https://github.com/vincemedia/brc-conformance**, under `brc-300/`, and the Implementations section links there. They are run against a checkout of this repository rather than shipping a copy of the document, so the two cannot drift:

```
python3 verify_vectors.py path/to/apps/0300.md
```

I have also written your reasoning into the document rather than just complying with it, because it is the better position and worth stating where an implementer will see it: a standards repository that accumulates executable code acquires a maintenance surface nobody signed up for, and a reader then has to work out whether the code or the prose is normative. The document says plainly that it is normative and the vectors are a check on it, and that where the two disagree the code has the bug.

Two notes for the record.

The conformance repository is under my account for now. If the BSVA would rather it sat in the organisation, or alongside the other reference implementations, say the word and I will move it and update the link.

And an apology: my earlier re-review request tagged `@deggen`, which is somebody else entirely. Sorry to whoever that is. You clearly saw it anyway, but I should have checked the handle.

On your other comment, about auditing and certifying wallet providers: I agree that is a whole BRC of its own, and section 9 item 3 now says as much in the document. Making catalog listing a precondition of bidding contains the malicious-vendor attack by relocating the trust into catalog governance, which BRC-137 only sketches, so the sponsored half of every chooser currently rests on a standard nobody has written. I would rather that gap be named in the text than discovered later. If the BSVA does pursue an audit-and-certify scheme, that is exactly the thing section 4.3.8 condition 10 would point at.

raspi-user:
I think this is a strong and much more serious proposal than a simple sponsored-wallet placement mechanism. The good parts are clear: the chooser requires real wallet choice, sponsored entries cannot supply their own card content, bounty amounts are hidden from the user-facing surface, settlement is pre-funded through BRC-227-style vouchers, and the document is unusually honest about what is verifiable versus merely asserted.

That said, I do not think this is ready to merge yet. My main concerns are:

1. **Catalog governance is now security-critical but remains out of scope.**

   The proposal correctly identifies that malicious wallet vendors can outbid honest ones, and that the catalog is the thing preventing paid placement from becoming a phishing/malware distribution channel. But the document then makes catalog listing a hard eligibility condition while leaving catalog governance largely unspecified. At minimum, BRC-300 should define the required catalog trust properties: signed entries, vendor identity key binding, install URL control, delisting/revocation freshness, conflict handling, and how exclusions propagate between catalogs. Otherwise the strongest security control in the design lives outside the standard.

2. **Settlement anchoring should be mandatory, not optional.**

   The bid set and delayed openings are doing real security work. But the anchor is currently only a `SHOULD`, and several claims are only verifiable "where anchored". That leaves room for equivocation: different records shown to different parties, missing losing bids, or rewritten bid sets. If auditability is part of the security model, the auction house should be required to publish an immutable commitment to the bid set at auction time, or the document should define an equivalent transparency mechanism.

3. **Claim timing versus delayed publication needs tightening.**

   Section 6.5 requires the sponsor to reject a claim unless the settled record names that sponsor, but section 8.4 delays publication of opened commitments until at least `max(maturityBlocks, 144)`. The spec needs to state precisely what record the sponsor relies on during claim validation, when it becomes available, and whether that record is private, public, anchored, or only later opened. Otherwise conforming implementations may disagree on whether a valid claim can be processed before the public delayed record appears.

4. **Silent automatic claiming is a significant privacy/consent decision.**

   The document is explicit that the newcomer's identity key is disclosed to the sponsor at claim, potentially without prompting. That may be an intentional UX tradeoff, but the chooser/card disclosure should say plainly that sponsored funding may result in the wallet identity key being disclosed to the sponsor. The current disclosure focuses on sponsorship mechanics, not enough on the identity/privacy consequence before the user chooses.

5. **The campaign still duplicates wallet-card data.**

   The proposal says cards must be rendered from the catalog, not sponsor-supplied content, which is good. But the campaign object still contains a `wallet` object with name/vendor/icon/description/install fields. That creates avoidable drift and equality/canonicalization problems. I would prefer the campaign reference a catalog entry/vendor key/platform target only, with all user-facing card content sourced exclusively from the catalog.

6. **Several deterministic details need to be specified more tightly.**

   Examples: the version model mixes numeric `version: 1` objects with "major component" language and manifest `"1.0"`; the tie-break rule needs an exact canonical hash input; predicate grammar needs clearer normalization for country/language/platform values, recursion/depth limits, duplicate terms, and BRC-52/BAP field/trust-anchor handling.

Overall: I support the direction, but I would treat this as **needs revision before merge**. The proposal is trying to standardize a security-sensitive onboarding market. The economics and UX are thoughtful, but the catalog, transparency, and privacy boundaries need to be nailed down before this becomes normative.

PR #227, opened 2026-08-17
BRC-169: certificate custody, self-handle discovery, and reverse resolution

vincemedia:
Found while implementing BRC-169 handle support in [Free Radio](https://freeradio.bsvb.net/), a live BRC-169 application, against the HandCash registry and HandCash Desktop.

**One note on vehicle before the substance.** The repository README asks that substantial revisions go into a new extending standard rather than an edit, so as not to disrupt existing implementations. I have proposed this as an edit because section 4.6 is a clarification of something section 4.1 already implies rather than a new requirement, and because splitting three tightly coupled sections across two documents would make both harder to read. If you would rather see the new surface in a separate BRC extending 169, say so and I will restructure it: the content transfers unchanged.

## What is missing

**The document runs one way.** Grammar, resolution, caching and the client algorithm all take a handle and return an identity key. An application with a connected wallet has the opposite problem: it holds the key, has proven the user controls it, and wants the handle to show them. Nothing in the document answers that.

**And it defines a certificate without saying where the holder keeps it.** Section 4.1 requires a verifier to check the certificate, and closes by noting that certificates are exchanged and selectively revealed over BRC-103 and BRC-104. Both presuppose the holder still has it. Nothing said so, and no part of the document mentions `acquireCertificate`, `listCertificates`, `proveCertificate` or `relinquishCertificate`.

That gap has a consequence in production. A surveyed wallet performs forward resolution and claiming correctly, then stores `{handle, display, identityKey, claimedAt}` in its own storage and discards the certificate, so `listCertificates` has nothing to return. That record is not an attestation: it is the wallet's own unsigned claim about itself, it survives the spending of the revocation outpoint, and no third party can check it. Keeping the fields and dropping the signature keeps the part that is easy to fabricate.

The same wallet does expose the handle, through a vendor method restricted by an origin allowlist. The allowlist includes `localhost`, so an integration passes in development and returns 403 on deployment, which is the most expensive moment to find out.

I want to be clear that none of this is a criticism of that implementation. It satisfies every explicit MUST in section 4.1. The specification did not tell it where to put the certificate, so it put it nowhere.

## What this adds

**Section 4.6, custody.** The certificate is stored through BRC-100 `acquireCertificate`, relinquished on release per 4.3, and MUST NOT be replaced by a derived summary. A wallet MUST NOT expose handle data through a vendor method in place of the certificate surface, and MUST NOT gate that surface by origin allowlist: BRC-116 section 4.4 already permissions certificate access with the user deciding and the decision revocable, and an allowlist replaces that with the vendor's judgement while failing undiagnosably for anybody not on it. An allowlist containing a development origin is called out specifically.

**Section 5.8, the connected user's own handle.** Three paths, in order, stopping at the first that yields a verified binding:

1. `listCertificates` for the handle type, verified per 4.1 including `subject` equal to the key proven in session. No network, no disclosure. Works wherever 4.6 is satisfied.
2. `discoverByIdentityKey`.
3. The user supplies a handle; the client forward-resolves it per 5.7 and compares `identityKey` against the proven key, accepting only on exact match.

Path 3 is **verification, not trust**, and that distinction is why it is permitted rather than tolerated: the registry attests the binding, the challenge signature proves the key, and the client checks the two meet. The user is not asked to be believed. A new rule forbids the thing it replaces: a handle typed into a field and displayed unchecked is not identity, and where others can see it, it is impersonation with an extra step. Nothing in the document previously ruled that out.

**Section 5.9, primary designation.** A key may hold several bindings, and section 2.5 already contemplates organisations, machines and agents holding many. Which one to show is the holder's decision, expressed as a self-issued BRC-52 certificate with `subject` and `certifier` both the holder's key. It is a preference and not an attestation, MUST be ignored unless it names a binding the client independently verified, and where none is set the client presents the set and MUST NOT choose silently. A third derived type identifier is added to section 4.5 by the same construction as the existing two.

**Section 5.10, reverse resolution.** An optional endpoint advertised as `metanet.handles.reverse`, returning bindings with their certificates for a given key. **Discoverable by default**, because a handle exists to be addressed and a binding findable from only one end is a worse address. A holder opts out with a directive signed by the subject key, in the manner of the forwarding record of 4.3 and for the same reason: a host must not be able to fabricate either the hiding or the exposing of a binding it did not author. The control sits in the wallet, with the user. A host must not indicate that it withheld a result, since a response distinguishing "no such key" from "this key asked not to be found" answers the question it was refusing to answer.

## Security Considerations

Three paragraphs added: that reverse resolution completes the loop and turns a key seen in a payment into a human-readable name, with BRC-228 named for payers who want neither direction; that a wallet keeping the fields and discarding the signature has kept nothing; and that an origin allowlist over the certificate surface is a security decision taken away from the user.

## Also in this diff

Two stale links, flagged separately so they are not mistaken for part of the feature: BRC-190 was renumbered to BRC-146 on merge and this document still pointed at `../apps/0190.md`, which does not exist. One body reference and one References entry.

## Verification

`verify_appendix.py` still passes in full. No existing normative text was altered, no vector changed, and the three derived type identifiers reproduce from their strings. The new one is `base64(SHA-256("metanet-handles primary designation v1"))` = `myjPvkwiyYPEpN5LrLLePQFYA7R6ES8TqTFGOAYg/V0=`.

Existing implementations are unaffected in the forward direction. Section 4.6 does make a wallet that discards the certificate non-conforming, which is the point of raising it, and section 5.8 path 3 exists precisely so applications have a correct path to ship while wallets catch up.

### DIFF
--- a/peer-to-peer/0169.md
+++ b/peer-to-peer/0169.md
@@ -195,7 +195,7 @@ A handle can be released and reassigned to a different person. Address book entr
 
 #### 4.5 Certificate type identifiers
 
-A [BRC-52](./0052.md) `CertificateTypeID` is an opaque 32-byte value, and no BRC defines an authority that allocates one. Rather than wait for an allocation that has no allocator, this document **derives** its two type values from a name, so that any implementation reproduces them from this section alone:
+A [BRC-52](./0052.md) `CertificateTypeID` is an opaque 32-byte value, and no BRC defines an authority that allocates one. Rather than wait for an allocation that has no allocator, this document **derives** its type values from a name, so that any implementation reproduces them from this section alone:
 
 ```
 type = base64( SHA-256( ASCII derivation string ) )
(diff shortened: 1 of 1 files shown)


vincemedia:
Updated. Five corrections from a self-review, one of which was a factual error I would not have wanted a reviewer to find first.

**Section 5.8 claimed something false, twice.** Path 1 said "no network request, no disclosure to any host" and rule 4 said "paths 1 and 2 disclose nothing to anybody". Both are wrong, and the document contradicts them two sections earlier: verifying per 4.1 requires fetching the manifest at `fields.domain` to check the certifier key, and performing the revocation check of 4.2 against an overlay or index at the freshness that section mandates. Path 1 therefore makes at least two requests, and the revocation query tells whoever answers it exactly which certificate is being checked. Replaced with what is actually true: no path is free of traffic, what paths 1 and 2 avoid is telling the *ecosystem host* that this application is asking about this user, and clients should reuse cached manifests and revocation results across a session.

**`fields.handle` meant two different things.** Section 4.1 uses it for the bare handle with `fields.domain` alongside; my 5.9 used it for `handle@domain`. Same field name, two meanings, one document, so an implementer had to know the certificate type before it could parse the handle. The designation now uses the same two fields as 4.1, and 5.9 rule 7 says why.

**The new type identifier was asserted but not vectored.** Section 4.5 note 4 says the handle-certificate value is covered by the worked signature in A.3, which is the document's own standard for a derived type. **Appendix A.9** now works the designation end to end with a real RFC 6979 signature and a negative vector. It earns its place beyond arithmetic: because `subject` and `certifier` are the same key, an implementer will reasonably doubt BRC-52's signing parameters still apply, and A.9 shows the holder computing `holderPriv * G` and a verifier computing `1 * holderPub` and both landing on the same signing key. It also shows what the signature does *not* prove: substituting the handle fails verification, but a designation that verifies still says nothing about whether the binding is valid, which is why 5.9 rule 2 discards any designation naming a binding the client has not independently verified.

`verify_appendix.py` gains fourteen checks for A.9 and passes in full.

**Two smaller gaps.** Section 2.1's grammar allows `+tag` and 5.9 never said whether a designation may name a tagged handle; it must strip the tag, as resolution does. And 5.9 rule 4 required a client with no designation to "present the set and let the user choose", which a server-side application or an agent cannot do. The real requirement is never to select silently, so that stays a MUST NOT and presenting a chooser became a SHOULD.

**One over-strong requirement relaxed.** Section 5.8 said a client must stop at the first path yielding a verified binding, which also forbade continuing in order to discover a second binding the wallet does not hold. That is a legitimate thing for an organisation handle, so clients may now continue through the remaining paths while still not falling back while an earlier one has succeeded.

sirdeggen (approved):

PR #230, opened 2026-08-19
BRC-229: Wallet-Native Elliptic Curve Point Multiplication as a BRC-98 Module

connormurray2:
## Summary

Defines BRC-229 as the `ecpm` BRC-98 semantic module for wallet-native elliptic-curve point multiplication. Applications use the existing BRC-100 `getPublicKey` method; this proposal adds no BRC-100 method, call code, Wallet Wire message, or optional interface member.

The module supports commutative masking protocols such as mental poker and verifiable shuffles while keeping derived private scalars inside trusted wallet code.

## Protocol

```text
p ecpm <apply|remove> <pointHex> <logicalProtocolID>
```

- `apply` returns `d·P`.
- `remove` returns `d⁻¹·P`.
- The point and operation are excluded from the derived-key identity so both operations select the same scalar.
- The canonical BRC-42/43 namespace is `p ecpm <logicalProtocolID>`, with `keyID`, counterparty, security level, and privileged selection retained in their existing fields.

The logical protocol ID remains 5–273 ASCII characters, preserving the full BRC-43 namespace. The BRC-98 dispatch envelope is explicitly bounded at 353 characters for `apply` and 354 for `remove`, both within BRC-100's 400-character transport limit.

## Security and permissions

- Only `getPublicKey` is permitted under `p ecpm`; signing, HMAC, encryption, identity-key use, and linkage operations cannot reuse the ECPM scalar.
- Input points and public-key counterparties must use canonical lowercase 33-byte compressed secp256k1 encodings, including an x-coordinate range check before parser use.
- Ordinary and privileged grants are isolated; level-2 grants are counterparty-scoped, and privileged requests require an approved reason.
- Protocol-scoped grants and concurrent prompt deduplication support high-volume masking workflows.

## Conformance and implementation

The BRC includes apply, remove round-trip, independent-wallet commutativity, counterparty separation, invalid-encoding, and protocol-length boundary requirements.

The companion TypeScript implementation is [bsv-blockchain/ts-stack#488](https://github.com/bsv-blockchain/ts-stack/pull/488). The branch is reconciled with current `master`, all review conversations are resolved, and the proposal is ready for final maintainer review.


sirdeggen:
This should be reframed as a proposal to change the wallet interface rather than attempting to change BRC 100 itself, which is fixed by definition.

ty-everett:
I think the primitive is valuable, but I do not think it should add a new method or call code to BRC-100. This looks like exactly the kind of extension the BRC-98 P-protocol/module namespace was created to support.

BRC-100 is intentionally fixed. Making `multiplyPoint` optional avoids invalidating existing structural implementations, but it still expands the interface, wallet wire, substrates, clients, processors, and capability-detection machinery. It also creates a precedent where every new cryptographic primitive adds another optional method to the supposedly unchanging interface.

Instead, I suggest reframing BRC-229 as the definition of a BRC-98 module scheme, perhaps `ecpm`, and routing the operation through an existing BRC-100 method such as `getPublicKey`.

BRC-229 itself can define the `ecpm` scheme and clarify that **P-modules are semantic extension points, not merely alternate permission filters**. Within a declared `p <scheme>` namespace, a supporting wallet may assign scheme-specific semantics to existing methods. Unsupported wallets already have the correct behavior: they reject the P-protocol because they do not have the named module.

### Proposed application-facing form

For example:

```ts
await wallet.getPublicKey({
  protocolID: [
    2,
    `p ecpm apply ${pointHex} mental poker deal`
  ],
  keyID: 'deck-mask-1',
  counterparty: 'self',
  privileged: false,
  seekPermission: true
})
```

would return:

```ts
{
  publicKey: dP
}
```

To remove the transformation:

```ts
await wallet.getPublicKey({
  protocolID: [
    2,
    `p ecpm remove ${maskedPointHex} mental poker deal`
  ],
  keyID: 'deck-mask-1',
  counterparty: 'self',
  privileged: false,
  seekPermission: true
})
```

would return:

```ts
{
  publicKey: d⁻¹P
}
```

The concrete grammar could be:

```text
p ecpm <operation> <pointHex> <logicalProtocolID>
```

where:

- `<operation>` is `apply` or `remove`;
- `<pointHex>` is a canonical 33-byte compressed secp256k1 point;
- `<logicalProtocolID>` is the remaining protocol-specific string and may contain spaces;
- the BRC-43 security level remains the first element of the normal `protocolID` tuple;
- `keyID` remains the ordinary, separately supplied BRC-43 key ID;
- `counterparty` remains the ordinary, separately supplied BRC-43 counterparty;
- `privileged`, `privilegedReason`, and `seekPermission` retain their existing BRC-100 meanings.

If retaining the shorter form is preferable, `apply` could be the default:

```text
p ecpm <pointHex> <logicalProtocolID>
```

with an explicit form for removal. I slightly prefer always including `apply` or `remove`, because it makes parsing and auditing unambiguous.

The result still conforms to the existing `getPublicKey` return shape. A compressed public key is simply an encoded secp256k1 point, so `{ publicKey: PubKeyHex }` carries the result without a new type or wire representation.

### Internal derivation semantics

The point and operation must be excluded from the derived key’s identity.

Given:

```text
p ecpm apply <pointHex> <logicalProtocolID>
```

the module should derive the scalar using a canonical module-specific derivation namespace equivalent to:

```ts
protocolID: [
  securityLevel,
  `p ecpm ${logicalProtocolID}`
]
keyID
counterparty
privileged
```

It then computes:

```text
apply:   d · P
remove:  d⁻¹ · P
```

This distinction is load-bearing. If `<pointHex>` or `apply`/`remove` is included in the BRC-42 invoice used to derive `d`, the wallet derives a different scalar for every intermediate point. Masks would no longer commute, and applying and removing a mask would not select the same key.

The canonical derivation domain should nevertheless retain `p ecpm` so these keys cannot collide with ordinary signing, encryption, HMAC, payment, or application keys using the same logical protocol and key ID.

For privileged requests, the same operation should be performed using the wallet’s privileged derivation root, exactly as existing key-related methods already do. Applying and removing a privileged mask must use the same `privileged` value, counterparty, logical protocol, and key ID.

`identityKey: true` should be rejected. `forSelf` should either be rejected or required to be absent/false because the module already defines exactly which derived scalar is being used, and overloading `forSelf` would introduce needless ambiguity.

### Why this is an intended P-module use

The module system already routes P-protocol uses of `getPublicKey`, encryption, HMAC, and signature methods through scheme-specific request and response handlers. Those handlers may validate or transform the request and transform the response.

It would be useful for this BRC to state explicitly that “transform” includes scheme-defined method semantics. A P-module is therefore not limited to deciding whether a normal BRC-100 operation is authorized. It may define what an existing method means inside its reserved namespace, subject to the method’s existing transport shape.

That interpretation gives BRC-98 an important architectural purpose: it becomes the extension mechanism for wallet capabilities without changing BRC-100 itself. Otherwise the ecosystem will continue adding optional members and call codes whenever a new wallet-native operation is required.

The current `PermissionsModule` adapter may need a wallet-internal execution hook because request/response transformation alone cannot calculate \(dP\) from \(dG\) and arbitrary \(P\). That is an implementation concern rather than an application-interface change. A wallet could expose a restricted internal capability to the trusted module, such as:

```ts
multiplyDerivedPoint({
  protocolID,
  keyID,
  counterparty,
  privileged,
  point,
  invert
})
```

The module should receive this narrow capability, not unrestricted access to the wallet’s root key or raw `KeyDeriver`.

### This also fixes a key-isolation problem

The current security argument says that a key reachable through `multiplyPoint` never signs or encrypts. That is the desired rule, but ordinary BRC-43 derivation does not enforce it.

Without a module-specific namespace and method policy, an application can call `createSignature`, `encrypt`, `decrypt`, or HMAC methods with exactly the same:

```text
protocolID + keyID + counterparty + privileged
```

Those methods derive the same private key. This matters because \(dQ\) is the ECDH shared secret for derived private key \(d\) and point \(Q\). The proposal correctly identifies key reuse as the real security risk, but the proposed standalone method does not itself prevent that reuse.

An `ecpm` module can enforce the required key type:

- permit only the declared EC point operations;
- reject signing under `p ecpm`;
- reject ordinary BRC-2 encryption/decryption under `p ecpm`;
- reject HMAC operations under `p ecpm`;
- reject linkage revelation unless BRC-229 deliberately defines it;
- never use the EC point multiplication key for spending or identity functions.

Thus the module does more than preserve interface compatibility: it makes the proposal’s “this key is only used for point multiplication” security invariant enforceable.

### Clarification concerning `getPublicKey`

The current text says that no companion derivation method is required because `getPublicKey` returns \(dG\) for the same derivation arguments.

That needs qualification when an explicit counterparty is supplied. In the current BRC-42 implementation, normal `getPublicKey` behavior may return the counterparty’s derived public key:

```text
P + hG
```

rather than the public key corresponding to the wallet’s derived private scalar:

```text
(x + h)G = dG
```

The latter ordinarily requires self-derivation or `forSelf: true`.

The module proposal avoids that ambiguity. Under `p ecpm`, the input point is explicit and the result is defined directly:

```text
apply(P)  = dP
remove(P) = d⁻¹P
```

Calling `apply` with the generator \(G\) naturally returns \(dG\), so no additional method is needed.

### Why counterparty naming alone is insufficient

BRC-43 lets the caller name \(P\) as the counterparty, but that does not make existing derivation operate with \(P\) as the base point.

Let the wallet identity scalar be \(x\), the named counterparty point be \(P\), and:

```text
h = HMAC(xP, invoice)
d = x + h
```

Existing BRC-42 public derivation can produce:

```text
dG = (x + h)G
```

or the counterparty child:

```text
P + hG
```

The required operation is:

```text
dP = (x + h)P = xP + hP
```

Naming `P` causes it to participate in the ECDH value used by the HMAC. It does not replace \(G\) with \(P\) in the final scalar multiplication.

If the caller knows \(P=pG\) and knows \(p\), it can compute:

```text
dP = p(dG)
```

That handles points whose discrete logarithm is already known, such as initial small card encodings. It does not handle arbitrary intermediate points after another participant has masked them, and it provides no way to calculate \(d^{-1}P\). The module primitive is therefore still necessary.

### Validation and permissions

The current validation requirements should be retained unchanged:

- exactly 33 bytes;
- compressed prefix `02` or `03`;
- canonical encoded coordinate checked before invoking a parser that may reduce modulo \(p\);
- valid secp256k1 point;
- not the point at infinity;
- reject a result at infinity.

The module should own the authorization semantics for its P-protocol, including originator identification, security levels, counterparty-specific grants, privileged access, `privilegedReason`, and `seekPermission`.

Because these protocols require hundreds of operations, the module should support protocol-level or grouped authorization instead of prompting for each point. This is another reason the P-module model is appropriate: the permission behavior belongs with the capability.

### Compatibility and feature detection

This design requires:

- no new BRC-100 member;
- no call code 29;
- no changes to Wallet Wire serialization;
- no changes to every substrate and structural implementer;
- no claim that old wallets have become non-conforming.

A wallet with the `ecpm` module processes the request. A wallet without it rejects `p ecpm` as an unsupported P-scheme, as BRC-98 already requires. Applications can handle that error and fall back to an application-held key or a trusted-dealer protocol.

In short, I recommend retaining the cryptographic and validation work in this proposal, but defining it as the `ecpm` BRC-98 module rather than adding `multiplyPoint` to BRC-100. This preserves BRC-100’s immutability, gives BRC-98 a concrete capability-extension role, provides enforceable key isolation, carries privileged and counterparty semantics through existing arguments, and avoids changing every wallet transport for one optional primitive.

wallet/0229.md
```diff
@@ -0,0 +1,211 @@
...
+- [BRC-42: BSV Key Derivation Scheme (BKDS)](../key-derivation/0042.md)
+- [BRC-43: Security Levels, Protocol IDs, Key IDs and Counterparties](../key-derivation/0043.md)
+- [BRC-44: Admin-reserved and Prohibited Key Derivation Protocols](../key-derivation/0044.md)
+- [BRC-72: Protecting BRC-69 Key Linkage Information in Transit](../key-derivation/0072.md)
+- [BRC-73: Group Permissions for App Access](./0073.md)
+- [BRC-98: P Protocols: Allowing Future Wallet Protocol Permission Schemes](./0098.md)
+- [BRC-100: Unified, Vendor-Neutral, Unchanging, and Open BSV Blockchain Standard Wallet-to-Application Interface](./0100.md)
+- [BRC-116: Wallet Permissions and Counterparty Trust](./0116.md)
+- <a name="footnote-1">1</a>: Barnett, A. and Smart, N. (2003). Mental Poker Revisited. Cryptography and Coding, LNCS 2898, pp. 370–383.
+- <a name="footnote-2">2</a>: Biehl, I., Meyer, B. and Müller, V. (2000). Differential Fault Attacks on Elliptic Curve Cryptosystems. CRYPTO 2000, LNCS 1880, pp. 131–146. (Invalid-curve attacks.)
```
ty-everett, line 211:
```suggestion
# BRC-229: Wallet-Native Elliptic Curve Point Multiplication as a BRC-98 Module

Connor Murray (connor.murray@bsvassociation.org)

## Abstract

This proposal defines `ecpm`, a [BRC-98](./0098.md) permission-module scheme for applying a wallet-derived secp256k1 scalar to an arbitrary caller-supplied curve point, or removing that scalar by applying its modular inverse.

The operation is carried over the existing [BRC-100](./0100.md) `getPublicKey` method. Inside the reserved `p ecpm` namespace, a supporting wallet interprets the method semantically as elliptic-curve point multiplication and returns the normal `{ publicKey }` result. No method, call code, Wallet Wire message, or optional member is added to BRC-100.

The scheme enables commutative-masking protocols such as mental poker and verifiable shuffles while keeping the derived scalar inside the wallet. Wallets that do not install the module retain BRC-98's required behavior and reject the reserved protocol.

## Motivation

A class of multi-party protocols depends on the commutativity of scalar multiplication:

```
a·(b·P) = b·(a·P)
```

Barnett-Smart mental poker is a representative use. Each participant applies a secret scalar to encoded card points, and later removes that scalar in any order. The application needs `d·P` and `d⁻¹·P` for an arbitrary point `P`, but it must not learn `d`.

Ordinary BRC-100 `getPublicKey` does not supply this operation. It performs BRC-42/43 child-key derivation and returns a derived public key. Naming `P` as the BRC-43 counterparty influences the child derivation, but does not replace the generator with `P`. Depending on `forSelf`, the result is the wallet's or counterparty's derived child public key, not the wallet's derived private scalar multiplied by the caller's point.

This distinction matters when `P` is an intermediate masked point whose discrete logarithm is intentionally unknown. From `d·G` and `P`, an application cannot compute `d·P` without solving a discrete logarithm or already knowing the scalar of `P`.

Existing key-linkage and encryption methods can be composed to emulate some forward-multiplication cases, but that is not an equivalent contract. It couples a point operation to linkage disclosure and encryption semantics, does not provide multiplication by `d⁻¹`, and grants permissions for capabilities the application did not mean to request.

The missing behavior therefore cannot be obtained from pure BRC-43 naming alone. It requires a wallet-side semantic operation, but it does not require changing BRC-100.

## Relationship to BRC-98

BRC-98 reserves protocol identifiers beginning with `p `, requires unsupported wallets to reject them, and permits a supported scheme to define its own rules for permitted operations, key IDs, counterparties, permission attributes, and execution.

The `ecpm` scheme uses that reserved dispatch point to specialize the meaning of an existing BRC-100 method. This proposal does not amend BRC-98. A clarification in discussion of BRC-98 may record that semantic specialization of an existing method inside a supported module namespace is an intended use of the module system; a separate amendment to BRC-98 is not required.

The semantic boundary is safe for compatibility:

1. Outside `p ecpm`, `getPublicKey` retains its ordinary BRC-100 meaning.
2. Inside `p ecpm`, a wallet either implements this scheme or rejects the request as BRC-98 already requires.
3. Applications never infer support from a new optional wallet member. They use a configured supporting wallet or handle the standard unsupported-scheme error.

## Specification

### Scheme and method

The BRC-98 scheme ID is:

```
ecpm
```

A conforming implementation MUST accept this scheme only through the existing BRC-100 `getPublicKey` method. It MUST reject every other BRC-100 method requested under `p ecpm`. This prevents the ECPM scalar from being reused for signing, HMAC, encryption, or another cryptographic purpose.

The protocol-name component of `GetPublicKeyArgs.protocolID` has this grammar:

```
p ecpm <operation> <pointHex> <logicalProtocolID>
```

where:

- `operation` is exactly `apply` or `remove`;
- `pointHex` is a lowercase, 66-character compressed secp256k1 public-key encoding;
- `logicalProtocolID` is the application protocol whose ECPM key universe is being requested.

The outer BRC-100 call remains:

```ts
wallet.getPublicKey(
  {
    protocolID: [
      securityLevel,
      `p ecpm ${operation} ${pointHex} ${logicalProtocolID}`
    ],
    keyID,
    counterparty,
    privileged,
    privilegedReason,
    seekPermission
  },
  originator
)
```

`keyID`, `counterparty`, `privileged`, `privilegedReason`, `seekPermission`, and `originator` retain their existing BRC-100 meanings and locations. They MUST NOT be duplicated or encoded into the protocol string.

`counterparty` defaults to `self` under the existing method rules. `identityKey: true` is prohibited. `forSelf` MUST be absent or false; it does not alter ECPM semantics.

### Logical protocol ID

The logical protocol ID MUST:

- contain only lowercase ASCII letters, numbers, and single spaces;
- contain no leading, trailing, or repeated spaces;
- be at least 5 characters and no more than 273 characters; and
- not end with ` protocol`.

The 273-character ceiling ensures that the canonical derivation protocol `p ecpm <logicalProtocolID>` remains within BRC-43's 280-character protocol-ID limit.

The complete outer `p ecpm` protocol string also MUST fit the active BRC-100 protocol-string limit. A key ID MUST satisfy the active BRC-100 `getPublicKey` key-ID limit.

### Canonical scalar derivation

Let:

- `s` be the security level in the outer `protocolID` tuple;
- `L` be `logicalProtocolID`;
- `K` be the separately supplied `keyID`; and
- `C` be the separately supplied `counterparty`, defaulting to `self`.

The module MUST derive a nonzero private scalar `d` using BRC-42/43 with:

```
protocolID  = [s, "p ecpm " + L]
keyID      = K
counterparty = C
```

The point and operation MUST NOT form part of this derivation identity.

This omission is load-bearing. If the point were part of the invoice, every point would select a different scalar. If the operation were part of the invoice, `remove` would select a different scalar from `apply`. In either case round-trip removal and commutativity would fail.

The `p ecpm ` prefix in the canonical derivation protocol isolates ECPM scalars from ordinary application keys using the same logical protocol and key ID.

### Operation

After validating the point as `P`:

- `apply` returns `d·P`;
- `remove` returns `d⁻¹·P`, where the inverse is computed modulo the secp256k1 group order `n`.

The successful result MUST use the ordinary `getPublicKey` result shape:

```ts
{ publicKey: PubKeyHex }
```

The returned value MUST be the lowercase compressed encoding of the resulting point.

For identical security level, logical protocol ID, key ID, counterparty, and root-key selection:

```
remove(apply(P)) = P
```

Implementations MUST reject, rather than encode, a result at infinity.

### Privileged keys

The existing `privileged` and `privilegedReason` fields select privileged-key behavior.

When `privileged` is absent or false, the module derives `d` from the wallet's ordinary BRC-42/43 root.

When `privileged` is true:

1. `privilegedReason` MUST be present and satisfy the existing BRC-100 description constraint.
2. The wallet MUST obtain explicit authorization even at security level 0.
3. The wallet MUST derive `d` from its privileged key material using the same canonical ECPM derivation tuple.
4. The privileged root or key-derivation capability MUST NOT be exposed to the application.
5. The wallet MUST fail closed if privileged derivation is unavailable.

A wallet implementation SHOULD request access to privileged material only after the operation and reason have been authorized, and SHOULD release that access according to its existing privileged-key policy.

### Permissions

The module applies BRC-43 permission semantics to the logical ECPM protocol:

- an ordinary security-level-0 request does not require a prompt;
- a security-level-1 grant is scoped to the originator, logical protocol ID, and ordinary-versus-privileged root selection;
- a security-level-2 grant additionally distinguishes the counterparty; and
- ordinary and privileged grants MUST NOT satisfy one another.

A wallet MAY cache a successful grant according to its normal permission lifecycle. The point, operation, and key ID SHOULD be displayed or made available to the authorization UI, but they do not change the BRC-43 protocol-level grant scope.

If permission is required and no applicable grant exists:

- `seekPermission: false` MUST fail without prompting;
- otherwise the wallet MAY seek authorization through its normal trusted UI.

Concurrent equivalent requests SHOULD share one pending authorization decision so that high-volume protocols cannot produce duplicate prompts.

### Point validation

Before multiplication, a conforming implementation MUST reject the input unless all of these conditions hold:

1. The encoding is exactly 33 bytes represented by 66 lowercase hexadecimal characters.
2. The first byte is `02` or `03`.
3. The encoded x-coordinate is less than the secp256k1 field prime `p`.
4. The encoding decodes to a point satisfying `y² = x³ + 7 (mod p)`.
5. The point is finite.

The x-coordinate range check MUST occur before a parser that reduces coordinates modulo `p`. Some curve parsers accept `02` followed by 32 `ff` bytes, reduce the x-coordinate, and then report the resulting different point as valid.

The same canonical validation MUST be applied when `counterparty` is supplied as a public key. The special values `self` and `anyone` remain valid.

### Errors

A conforming wallet MUST fail without performing multiplication when:

- the installed method is not `getPublicKey`;
- the protocol grammar or any field constraint is invalid;
- identity-key or `forSelf: true` behavior is requested;
- the point or public-key counterparty is invalid;
- authorization is unavailable or denied;
- privileged derivation is requested but unavailable; or
- the operation would return infinity.

Errors SHOULD identify the unsupported scheme, invalid field, or denied capability without revealing key material.

## TypeScript reference module

The TypeScript reference implementation is proposed in [bsv-blockchain/ts-stack#488](https://github.com/bsv-blockchain/ts-stack/pull/488) as the installable `@bsv/ecpm-permission-module` package.

It extends the Wallet Toolbox permission-module interface with an optional semantic handler:

```ts
handleRequest?: (
  request: { method: string; args: object; originator: string },
  next: (args: object) => Promise<unknown>
) => Promise<unknown>
```

A semantic handler may return a conforming BRC-100 result directly, or invoke `next` at most once to use the underlying wallet method. Existing `onRequest` and `onResponse` transformation modules remain source-compatible.

The ECPM module returns the result directly and never forwards its request to ordinary `getPublicKey`; forwarding would derive `d·G` or a BRC-42 child public key, not `d·P`.

A wallet host installs it alongside other BRC-98 modules under the `ecpm` scheme and supplies:

- its ordinary BRC-42/43 key deriver;
- its trusted authorization callback; and
- optionally, a privileged-key-deriver provider.

The key derivers and derived scalars remain inside trusted wallet/module code and are never included in the BRC-100 response.

## Security

### Key isolation

For a counterparty point `Q`, `d·Q` is an ECDH shared secret. Reusing an identity, spending, signing, HMAC, or encryption key for ECPM could therefore disclose a capability belonging to another protocol.

For this reason, the canonical `p ecpm <logicalProtocolID>` derivation namespace and the method restriction are mandatory. A wallet MUST NOT substitute its identity key, spending key, or a key from an ordinary non-ECPM protocol.

### Chosen-point requests

For a valid prime-order secp256k1 point `P`, observing `P` and `d·P` does not reveal `d` without solving the elliptic-curve discrete logarithm problem. An adaptively selected valid point does not weaken this assumption in the prime-order group.

That argument depends on strict validation. Accepting a point on another curve or a non-canonical encoding can create small-subgroup or invalid-curve attacks that leak information about `d`. Implementations must validate the encoded point before multiplication.

### Permission and denial-of-service considerations

Mental-poker and shuffle protocols can require hundreds of operations, so prompting once per point is impractical. Protocol-scoped grants allow the intended throughput. Wallets SHOULD still bound request concurrency and resource use, and SHOULD make the logical protocol, originator, operation, key ID, counterparty, point, and privileged reason available to trusted policy code.

### Capability containment

The public application interface exposes only `getPublicKey` and receives only a compressed public point. The semantic module is trusted wallet code: it may use internal key-derivation capabilities, but it MUST NOT return those capabilities, a private scalar, or privileged root material.

## Why pure BRC-43 is insufficient

BRC-43 can name the counterparty and thereby determine which shared derivation universe is used. It cannot name a replacement generator for the requested group operation.

If the arbitrary point is supplied as `counterparty`, ordinary `getPublicKey` performs BRC-42 child derivation relative to that counterparty. It does not compute the wallet's derived private scalar times that point. If `getPublicKey` returns `d·G`, combining it with an arbitrary `P` still does not yield `d·P` unless the application knows the discrete logarithm of `P`. Intermediate points in a commutative-masking protocol are constructed specifically so that no one knows that logarithm.

Pure BRC-43 therefore provides the scalar namespace and permission vocabulary used by this proposal, but not the required group operation. BRC-98 supplies the missing semantic dispatch without changing BRC-100.

## References

- [BRC-2: Data Encryption and Decryption](./0002.md)
- [BRC-42: BSV Key Derivation Scheme](../key-derivation/0042.md)
- [BRC-43: Security Levels, Protocol IDs, Key IDs and Counterparties](../key-derivation/0043.md)
- [BRC-44: Admin-reserved and Prohibited Key Derivation Protocols](../key-derivation/0044.md)
- [BRC-73: Group Permissions for App Access](./0073.md)
- [BRC-98: P Protocols](./0098.md)
- [BRC-100: Unified Wallet-to-Application Interface](./0100.md)
- [BRC-116: Wallet Permissions and Counterparty Trust](./0116.md)
```

ty-everett (commented):
Suggested full replacement: specify BRC-229 as the `ecpm` BRC-98 semantic module rather than adding an optional BRC-100 method. This keeps BRC-100 and its Wallet Wire unchanged, retains key ID/counterparty/privileged controls in their existing fields, defines apply/remove over one canonical derived scalar, and explains precisely why naming the point as a pure BRC-43 counterparty is not equivalent. The companion reference implementation has been revised in bsv-blockchain/ts-stack#488.

raspi-user (commented):
I reviewed BRC-229. I think the underlying primitive is valuable: wallet-held scalar multiplication over arbitrary secp256k1 points is a real missing capability for commutative masking protocols. However, I would not merge this draft in its current form.

The main issue is architectural. The proposal adds `multiplyPoint` as an optional method on BRC-100. Making it optional avoids retroactively invalidating existing wallets, but it still changes the supposedly fixed BRC-100 application interface, wire surface, substrate implementations, feature-detection rules, and client adapters. This looks like exactly the case BRC-98 was intended to handle. I think BRC-229 should define a `p ecpm` semantic module instead of adding a new BRC-100 method.

That also fixes the main security isolation problem. The current draft says the key used for point multiplication must be BRC-42/43-derived and must not be an identity/spending key. That is necessary, but not sufficient. If an application can use the same `protocolID`, `keyID`, `counterparty`, and `privileged` values with `createSignature`, `encrypt`, `decrypt`, HMAC, or linkage methods, then the supposedly point-multiplication-only key is not actually isolated. A `p ecpm` namespace can make the key universe explicit and reject non-ECPM operations under that namespace.

There is also ambiguity around the statement that no companion derivation method is needed because `getPublicKey` already returns `d·G` for the same derivation arguments. With counterparty derivation, existing BRC-42/BRC-100 public-key semantics can return the counterparty-derived public key rather than "the wallet's derived scalar times G" in the sense needed by this operation. The spec should define the scalar derivation domain directly and avoid depending on an ambiguous analogy with `getPublicKey`.

Specific changes I would want before merge:

1. Reframe BRC-229 as a BRC-98 module, probably `p ecpm`, rather than a new optional BRC-100 method.
2. Define exact module grammar, e.g. `p ecpm apply <pointHex> <logicalProtocolID>` and `p ecpm remove <pointHex> <logicalProtocolID>`, or another deterministic equivalent.
3. Specify that `apply` / `remove` and the supplied point are excluded from the scalar derivation identity, while the derivation namespace still includes `p ecpm`.
4. Normatively reject signing, encryption/decryption, HMAC, spending, identity-key use, and linkage revelation under the ECPM key universe unless a future BRC explicitly defines them.
5. Define permission grants precisely: originator, security level, logical protocol, key ID, counterparty, privileged flag, and whether apply/remove are separately grantable or one capability.
6. Replace "compressed DER" terminology with the exact point encoding, likely canonical 33-byte compressed SEC1 secp256k1 public-key encoding.
7. Define exact validation and error behaviour: x-coordinate range before parser use, valid curve point, not infinity, result not infinity, scalar zero/inversion failure, malformed hex, wrong length, wrong prefix, and unsupported module.
8. Add conformance vectors for apply, remove, round-trip, multi-party commutativity, counterparty separation, privileged/non-privileged separation, invalid encodings, out-of-range x, and unsupported module behaviour.
9. Update the implementation references. The current draft points to `ts-stack#487`, which is closed/superseded; the active direction appears to be `ts-stack#488` using the BRC-98 ECPM module.

Overall: the cryptographic primitive is worth standardising, but it should be standardised as a wallet capability module with enforced key isolation, not as a new optional BRC-100 method. The current draft should be revised to match the ECPM module architecture before merge.

connormurray2:
The draft has been revised to the ECPM module architecture — I accepted @ty-everett's full-file replacement reframing BRC-229 as the `p ecpm` BRC-98 scheme over `getPublicKey`. (If you read the branch on Aug 30: GitHub truncated the applied suggestion at its first nested code fence, leaving the file ending mid-Motivation. The full text is restored as of ebf0607.)

Against your list:

1. **Reframed as a BRC-98 module** — the `p ecpm` scheme, carried over `getPublicKey`. No method, call code, or wire message is added to BRC-100.
2. **Exact grammar** — `p ecpm <operation> <pointHex> <logicalProtocolID>`; the operation is always explicit (`apply` or `remove`).
3. **Derivation identity** — the canonical derivation is `[s, "p ecpm " + logicalProtocolID]` with keyID and counterparty in their ordinary fields; the operation and point are normatively excluded from the derivation identity, and the spec explains why that omission is load-bearing.
4. **Key-universe isolation** — only `getPublicKey` is accepted under `p ecpm`; every other BRC-100 method — signing, encryption, HMAC, linkage — is rejected.
5. **Permission grants** — scoped to originator + logical protocol (+ counterparty at level 2); ordinary and privileged grants never satisfy one another; apply/remove share one capability, with point and operation surfaced to the authorization UI.
6. **Encoding** — pinned to exactly 33 bytes, 66 lowercase hex characters, `02`/`03` prefix.
7. **Validation and errors** — normative, including the x-coordinate range check before any reducing parser.
8. **Conformance vectors** — added in 8226597: apply, remove round-trip, two-wallet commutativity, counterparty separation, and rejected encodings (including the out-of-range x case). Generated with go-sdk's BRC-42 `DeriveChild` over the canonical derivation.
9. **Implementation reference** — now points to bsv-blockchain/ts-stack#488.

One open question: the 273-character logical-protocol ceiling assumes only the *derivation* string must fit BRC-43's 280-character limit, but the outer string carries ~80 characters of `p ecpm apply <pointHex> ` overhead. If the 280 limit applies to the outer protocol string too, the effective ceiling is 199. Either the stated ceiling should drop or the spec should exempt the outer `p`-protocol string explicitly.

ty-everett:
Yeah, @connormurray2 I agree that last point can get a little ugly.

I'm leaning towards just thinking that the spec should carve out a wider limit for its prepended metadata, and declare a slightly oversized but still bounded character limit.

That way the spec can work with all otherwise-valid logical protocol IDs, and still specify the other fields.

If we're all good with that, I'll revise the BRC and branch accordingly, and then we can get both merged and published.

ty-everett:
Thanks @raspi-user for the thorough review, and @connormurray2 for working through the redesign. The proposal now defines ECPM as a BRC-98 semantic module, and the final character-limit item is resolved: logical protocol IDs retain the full 273-character allowance, while the outer dispatch envelope is explicitly bounded at 353 characters for `apply` and 354 for `remove`.

The branch is reconciled with current `master`, the index and PR description match the module architecture, all review conversations are resolved, and this PR is now ready for final maintainer review.

ty-everett (approved):

PR #251, opened 2026-09-16
BRC-178: Federated Identity Bootstrap Proxy for Application Authorization

GenericCPU:
## Summary

Defines a vendor-neutral HTTP boundary for applications to bootstrap application authorization through a shared federated identity provider without receiving provider OAuth credentials or tokens.

## Deployed profiles

- HandCash Market proxy
- Soundbase relying application
- 402AI preproduction relying application

## Security properties

- proxy-relative return paths and registered final callback origins
- single-use state bound to an ephemeral authorization key
- no provider tokens exposed to the relying application
- ordinary authorization/consent/2FA policy remains in force

**Status:** Draft / Beta


GenericCPU:
Withdrawing this draft. HandCash is documenting the Google/Auth0 app proxy as a product feature of HandCash Connect rather than as a BRC — it bootstraps a Connect authorization and is not part of the BRC-100 wallet interface, so a standards track entry isn't the right home for it. Apologies for the churn, and thanks for the look.

